PR #11, opened 2015-08-28
FIX: Allow writing datasets without template

ischwabacher:
Use default `IJK_TO_DICOM_REAL` if no template header is passed to `write.c.AFNI` instead of erroring on missing variable.

Reference: [http://afni.nimh.nih.gov/afni/community/board/read.php?1,149233,149240#msg-149240]


ischwabacher:
Already fixed.

PR #12, opened 2015-10-06
Matlab library: replace logical bit-wise operators 

nno:
Make the Matlab library more compatible with GNU Octave by replacing logical bit-wise operators ('&' and '|') by the short-circuit versions ('&&' and '||').

### DIFF
--- a/src/matlab/CheckBrikHEAD.m
+++ b/src/matlab/CheckBrikHEAD.m
@@ -81,7 +81,7 @@ for (ir = 1:1:N_Rules),
 					err = 1; ErrMessage = sprintf('Error %s: Field %s type (string or numerical) is wrong.', FuncName, Rules(ir).Name);warndlg(ErrMessage);return;
 				end
 			else %a number, verify type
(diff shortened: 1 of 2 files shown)


afni-rickr:
Ick.  That is a nice change in any case, thanks!  :)

nno:
Thanks @afni-rickr for merging!
There are other functions in the AFNI Matlab library that use logical bit-wise operators where short-circuit versions would be better - hopefully soon I will be able to find time to work on those and send a new Pull Request.

afni-rickr:
I suspect that code is ancient, and our principle Matlab coder is no longer with the group.  We might not maintain this much in the future.  Thanks again!

nno:
Sorry to hear that Ziad is no longer with the group. In any case I like the Matlab library; there is room for some minor improvements, such as use of logical operations, and some refactoring to use more Matlab-style solutions than C-style solutions, but nothing of that is critical. I hope to be able to make more contributions in the future.

PR #18, opened 2015-12-07
BUILD: Make build process slightly more portable (?)

ischwabacher:
This PR contains modifications I made to get AFNI building on our machine. I've been rebasing these changes over upstream/master on a regular basis, but since I tried to make sure that everything would be as portable as possible, it makes sense to see if these work on the original build machines.

WARNING: This PR has been **rebased** (as of 2015-12-07 15:33:00-0600)!


afni-rickr:
Would you clarify the reason for removing CCOLD from Makefile.xorg7_64, for example?  From what did you conclude it was optimizing out functions?  The xorg7 and openmp builds are still active, so I want to be careful.  If I get some time, I will do a build of your fork on those 4 platforms, and see if this fixes the observed incorrect computations (e.g. convert F to p, see 19 May 2010 update).  Please send me direct email if you would like to discuss this in more detail.

ischwabacher:
It looks like this is not likely to be a net improvement to AFNI. Closing.

PR #29, opened 2016-04-28
Converted top level README into README.rst and added a travis badge

yarikoptic:
I know that aesthetics are subjective, but I thought this one would be better. See how it looks by visiting my clone: https://github.com/yarikoptic/afni-1 and also noting that nice travis badge at the bottom ;)


yarikoptic:
chose .rst instead of .md since you use rst for docs

yarikoptic:
OMG -- I haven't broken anything! ;)

nno:
@yarikoptic thanks for adding the badge - but wouldn't it make more sense to have it at the top? It does not take up much space, and that way you can see directly (without scrolling) whether the build passes.

yarikoptic:
done @nno

PR #30, opened 2016-04-28
RF: remove -lXp (I do not think it is used, obsolete, removed from debian) from Makefile*ubuntu*

yarikoptic:
I think we never linked against in our NeuroDebian builds...  do you actually need it?  sending to travis to check ;)


yarikoptic:
any comments/ideas @afni-rickr  and @afni-rwcox holding you away from pressing the green button? ;-)

afni-rickr:
This is something that I would want to verify locally.

One thing to note about those top-level Makefiles: they were not written for general use, but for creating distribution binaries on very specific machines.  It is on those machines that I would like to test.

On that note, we would only change active Makefiles or the ones under other_builds, not alter historic Makefiles.  Maybe I should move the old ones to a sub-directory, we'll see.

At any rate, I will start looking into this change, but will do so slowly.
Perhaps the best places to start are for the openmp and ubuntu Makefiles, since those are used on systems that seem to not come with libXp anymore.

Does that seem reasonable?

yarikoptic:
whatever floats your boat @afni-rickr .  as long as there is activity/progress - I would be happy ;-)  I would also be interested to discover which systems still alive need that libXp.  And FWIW, you (if you use bash or zsh) might benefit from this helper I use:  http://git.onerussian.com/?p=etc/bash.git;a=blob;f=.bash/bashrc/30_aliases_sh;hb=HEAD#l857

so I would do `ghpr co-head https://github.com/afni/afni/pull/30` to get a pr-30 branch which would be with the proposed PR state.  or use "co" instead to get a state with merged master.  This way I can review/check PRs quickly on my system

afni-rickr:
Yes, libXp is used on our older build servers (for linux_xorg7 and linux_openmp) as a dependency of openmotif.
It does not seem to be needed for the more modern linux_fedora_21 package.

Maybe we should create a new build for Ubuntu 14.04 (which would hopefully work on 16.04, too).  That might resolve the problem, since those systems currently rely on linux_openmp.

yarikoptic:
@afni-rickr please advice on what should I do so I could also push #36 forward.  I guess I could abandon all changes in the files you would like to preserve or explicitly target only the ones I guess on travis (OS_notes.linux_ubuntu_12_64).

afni-rickr:
Actually tested Fedora 21 and Ubuntu 14.04.  Fedora 21 does still depend on libXp.  Only Ubuntu systems (tested 14.04) do not seem to use it, apparently due to a different Motif package.  So at this point, only the 3 other_builds/*ubuntu* files seem to be candidates for a change, but I do not have an Ubuntu 12 system to verify even that much.
Are you actually building on Ubuntu 12, or is it 14 or 16?  If you are not testing on 12, then I would be more inclined to simply make new Ubuntu 14 files that do not include libXp (easier than creating a new system).

yarikoptic:
Travis based system iirc Ubuntu precise 12.04

afni-rickr:
Given that, changing the 3 Ubuntu 12 files seems reasonable.  I plan to make a distribution package that does not depend on Xp (for Ubuntu systems) sometime soon (so within 43 years), and will make a more formal Makefile at that point.

yarikoptic:
On Wed, 18 May 2016, afni-rickr wrote:

>    Given that, changing the 3 Ubuntu 12 files seems reasonable. I plan to
>    make a distribution package that does not depend on Xp (for Ubuntu
>    systems) sometime soon (so within 43 years), and will make a more formal
>    Makefile at that point.

I am not that optimistic then to see the point when I should proceed
forward... or could you provide me intermediate guidance? ;)  primarily
I just want to see #36 merged ;)  but it is based on this one to be able
to build it iirc (sorry for impaired long term memory ;) )

afni-rickr:
It seems that there are only 7 files in that pull request that should be changed, 4 of the new ones that you created: .travis.yml, CMakeLists.txt, tests/CMakeLists.txt, tests/xvfb-driver, plus 3 of our old ones: src/other_builds/*ubuntu*.
So what is the easiest way to revert the other 56 files?
Can this commit be removed (da7b995e5bbcf60700b873e111d9249dc75ce4ae)?

yarikoptic:
I will 'rewrite' that commit to modify only 3 files you have mentioned, and then will push to this PR.  whenever we are satisfied with them here, I will rebase that PR against this state

yarikoptic:
and I guess I should leave the `-lXp` in other Makefiles such as
```
$> git diff src/rwcox/Makefile src/XmHTML/Makefile
diff --git a/src/XmHTML/Makefile b/src/XmHTML/Makefile
index 560e652..ef7dabf 100644
--- a/src/XmHTML/Makefile
+++ b/src/XmHTML/Makefile
@@ -173,7 +173,7 @@ PLATFORMINC=-I../../include/XmHTML -I/usr/X11R6/include
 INCLUDES = $(LIBINC) $(PLATFORMINC) $(IMAGEINCLUDES) $(II)
 
 LIBDIR = -L/usr/X11R6/lib -L../../..
-LIBS   = -lXm -lXpm -lXmu -lXp -lXt -lXext -lX11 -lSM -lICE
+LIBS   = -lXm -lXpm -lXmu -lXt -lXext -lX11 -lSM -lICE
 
 # final line fed to the linker 
 LOADLIBES = $(LIBDIR) $(LIBS) $(IMAGELIBS)
diff --git a/src/rwcox/Makefile b/src/rwcox/Makefile
index 31e518d..9aaa81c 100644
--- a/src/rwcox/Makefile
+++ b/src/rwcox/Makefile
@@ -27,7 +27,7 @@ LINT = /usr/bin/lint -a -b -u -v -x $(IFLAGS) $(CCDEFS)
 INSTALLDIR = $(HOME)/abin
 LIBDIR = $(INSTALLDIR)
 
-LLIBS  = -lmri /usr/X11R6/lib/libXm.a -lXp -lXpm -lXext -lXmu -lXt -lX11 -lm  -ldl -lc
+LLIBS  = -lmri /usr/X11R6/lib/libXm.a -lXpm -lXext -lXmu -lXt -lX11 -lm  -ldl -lc
```
?

afni-rickr:
Yes, it would be good to leave it everywhere but in the Ubuntu files.

yarikoptic:
actually for the purpose of testing I guess I could even indeed just drop that commit iirc... let me try... I will just push changes to that branch and see how it goes

yarikoptic:
So seems to be doing fine, and could be merged I guess

afni-rickr:
Great, thanks!

PR #36, opened 2016-04-29
NF: Extracted cmake configuration for running some rudimentary smoke tests

yarikoptic:
as I had it in the neurodebian package

This one sits on top of #30 which I had to have locally to make it all build.  So shouldn't be merged as is until decision on #30 is made (thus marked as WiP -- work in progress).  Let's now see if this all would work ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,16 @@ env:
 before_install:
     # install build-dependencies as the doc says
     - sudo sh src/other_builds/OS_notes.linux_ubuntu_12_64
+    # install packages needed for testing
+    - sudo apt-get install xvfb xauth libgl1-mesa-dri mesa-utils tcsh libjpeg-progs
 install:
     - cd src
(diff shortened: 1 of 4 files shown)


yarikoptic:
woohoo -- works from the first try!
Having this test would assure that SUMA built fine and runs `@DO.examples` successfully

I would be happy to tune it up the way you like it, but it could be a nice start for cmake-ification, although somewhat from the "back" ;)

yarikoptic:
ok -- picked up only testing related patches and placed them on top of current master -- let's see if testing goes smoothly. lXp issue would stay within #30 ;)

yarikoptic:
Yeay!  tests pass! whenever merged, I will adjust #30 to tune only ubuntu  make files and then we could see if it breaks any of our brand new tests somehow ;)

afni-rickr:
Great, thanks!

PR #39, opened 2016-06-27
Replace mpeg_encode with ffmpeg

mih:
Rational: AFNI relies on mpeg_encode for creating video clips. However, mpeg_encode is not maintained for years, has a problematic license/patent situation and is not available in any recent distribution. This patch replaces mpeg_encode with ffmpeg -- a more modern, maintained and available solution to achieve exactly the same result.
Originally written: 2009-10-29


afni-rickr:
Sorry for not getting to this yet.  I want to change this for either mpeg_encode or ffmpeg to be the backup of the other.  Many systems do not have ffmpeg.  And since this code merely deals with what is in the PATH, license issues are not really important (that can be a compile-time decision).

leej3:
Done. See 7d073c41fc68c05ad27533fec6d5ff0a95d044da

PR #40, opened 2016-06-27
call glutInit in SUMA_STANDALONE_INIT

mih:
Otherwise suma would crash with

  freeglut  ERROR:  Function <glutBitmapWidth> called without first calling 'glutInit'.

upon right click on a mesh

Originally written: 2012-11-29

### DIFF
--- a/src/SUMA/SUMA_DataSets.h
+++ b/src/SUMA/SUMA_DataSets.h
@@ -21,6 +21,7 @@
 	      }  \
          /* SUMAg_CF->scm = SUMA_Build_Color_maps();  require X connection*/\
       SUMA_ParseInput_basics_s (argv, argc);   \
+      glutInit(& argc, argv); \
    }
 #endif   
 


yarikoptic:
Dear @afni-rickr and @afni-rwcox ... do you think this fix is still pertinent or no longer (given its age, may be it was already handled...)?

afni-rickr:
No, we are just *really* slow to get to such things lately...

afni-rickr:
I took this out and will look into altering it.  It crashes on systems without a display, e.g. via ssh.

fenollp:
[This should help](http://stackoverflow.com/questions/6281998/can-i-run-glu-opengl-on-a-headless-server) ;)

yarikoptic:
Or it could be just conditioned on having DISPLAY env var set?

afni-rickr:
Yes, the headless rendering is a very useful link, but putting that call elsewhere or making it conditional seems appropriate in this case.  The user might be just getting the -help output, so it would be nice to have it later in the flow.  Of course, then it might vary per program.

PR #42, opened 2016-07-18
Added more centrality options and multi-scale entropy

ccraddock:
Added optimized functions for calculating:

* degree centrality
* eigenvector centrality
* local functional connectivity density
* multi-scale entropy
* SVM prediction when using linear models.


ccraddock:
Rick Reynolds emailed to request changes to this PR, will resubmit

PR #45, opened 2016-08-23
removed -V option from CCOLD line

andy5995:
The build failed at nifti/nifticdf/nifticdf.c because gcc was given the
-V option


afni-rickr:
The Makefiles are specific to our systems, and the ones that use those files do indeed need the -V option.
Expect to modify the Makefiles to suit your system peculiarities.  Consider starting with the Makefiles under other_builds instead, such as Makefile.linux_fedora_19_64 or Makefile.linux_ubuntu_12_64_OMP.  They even come with OS_notes files.

andy5995:
Understood. Thank you for the explanation.

PR #47, opened 2016-09-02
WiP RF: the same .travis.yml to build/test both upstream and debian (if debian/ is present)

yarikoptic:
In https://github.com/neurodebian/afni/pull/5 decided to make .travis.yml test building of the debian package (instead of native upstream build which would fail since some 3rd party code gets dropped for debian pkg). So I thought that may be you guys wouldn't mind if those scripts live straight in your afni code base so we don't carry that patch, thus needing to deal with conflicts etc.

I will also test if it works when building against debian branch (ie instead of https://github.com/neurodebian/afni/pull/5) before removing the WiP


yarikoptic:
travis failure seems to relate to upstream issue and travis does report fail to build condition for master:
https://travis-ci.org/afni/afni

afni-rickr:
That should hopefully be fixed now, thanks!

nno:
The current error is:
```
$ if [ -e debian ]; then tools/ci/debian_run_tests; else tools/ci/run_tests; fi
+mkdir ../testing
+cd ../testing
+cmake -DAFNI_BUILD_TESTS:BOOL=ON ..
CMake Error: The source directory "/home/travis/build/afni" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
The command "if [ -e debian ]; then tools/ci/debian_run_tests; else tools/ci/run_tests; fi" exited with 1.
```
Since current master contains CMakeLists.txt in the ``tests`` directory, maybe the PR should be updated to reflect that?

yarikoptic:
may be we could get back to it but later, and it would require rebase on top of "recently" rewritten history

PR #51, opened 2017-03-07
Updated source files to reflect bug correction. Now to use the matlab…

imnotamember:
… original form you must use the option '-legacy_transform 1'


imnotamember:
I didn't receive your earlier change today in my clone of the master branch. Would you prefer re-committing that on your end? Specifically, it is in "RVT_from_PeakFinder.py" on line 47.
`dm = min(len(r['p_trace']), len(r['p_trace']))`
should be:
`dm = min(len(r['p_trace']), len(r['n_trace']))`

afni-rickr:
That is great, thanks!
Sure, I will just re-commit the other little change.

PR #52, opened 2017-03-09
Added exception handling to all os.chmod calls.

DavidEWarrenPhD:
Added exception handling to all os.chmod calls (for filesystems that do not support chmod operations).


DavidEWarrenPhD:
Great, thanks!  This will make my life a little easier.   😄

PR #57, opened 2017-08-20
Make ctest run verbose

yarikoptic:
Just a wild guess while waking up

Issue #56, opened 2017-08-20
Failing ctest using Travis


nno:
(Issue was originally posted on the AFNI message board: https://afni.nimh.nih.gov/afni/community/board/read.php?1,155735,155735#msg-155735)

ctest shows failures on the travis ci environment (see [travis-ci.org]): 

23.28s$ ctest 
Test project /home/travis/build/afni/afni/testing 
Start 1: afni:start 
1/2 Test #1: afni:start ....................... Passed 5.98 sec 
Start 2: do.examples 
2/2 Test #2: do.examples ......................***Failed 17.24 sec 
50% tests passed, 1 tests failed out of 2 
Label Time Summary: 
GLX = 17.24 sec 
Total Test time (real) = 23.23 sec 
The following tests FAILED: 
2 - do.examples (Failed) 
Errors while running CTest 
The command "ctest" exited with 8. 

As for the underlying cause, I noticed that the issue seems to have started when travis changed from the older "Precise" to the newer "Trustry" environment. A temporary fix would thus be to explicitly set the environnment to be "Precise" as explained here: https://docs.travis-ci.com/user/precise-to-trusty-migration-guide/. This would require just adding a single line to .travis.yml, and then travis is again useful to catch compilation errors (amongst other things). 

For the long term it would be preferably to find out why the second do.examples fails and fix the underlying issue - that way the tests also work using Trusty, and use of travis to test AFNI is more "future-proof".

ping @yarikoptic: any ideas how we can diagnose the issue? Can we get more verbose output when running the ctest tests?

nno:
This was fixed in #58, so I'm closing the issue.

PR #58, opened 2017-08-22
Travis: set PATH variable so that afni binaries precede other binaries

nno:
This addresses failing travis tests due to using the wrong `count` program.

Addresses #56; see also AFNI message board discussion at https://afni.nimh.nih.gov/afni/community/board/read.php?1,155735,155735#msg-155735

Thanks to @yarikoptic for PR #57.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,6 @@ script:
     - export PATH=$PATH:$PWD/$FLAVOR  # to gain access to built binaries
     - mkdir ../testing; cd ../testing # from here will run tests
     - cmake -DAFNI_BUILD_TESTS:BOOL=ON ..
-    - ctest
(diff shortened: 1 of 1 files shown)


afni-rickr:
@nno, have you been able to verify where the other count program is coming from?
Thanks!

nno:
> have you been able to verify where the other count program is coming from?

@afni-rickr I'm trying that out at the moment in a temporary branch: https://github.com/nno/afni/tree/_debug/find_other_count

Output in https://travis-ci.org/nno/afni/builds/267200937

nno:
The other binary called count is in /usr/local/clang-3.5.0/bin/count
See https://travis-ci.org/nno/afni/builds/267200937

PR #60, opened 2017-08-27
WIP: improve cmake

tharvik:
Draft of cmake support for the whole project.
I tried to forward as most as possible the Makefile, with simplification and drop many weird construct.

As I have no knowledge of the use of the whole project, I don't really know how/what to test to see if everything is working, but the tests are passing. Also, I don't have Windows, so I couldn't test it there.

I guess I forgot some output and misunderstood some part of the Makefile, YMMV, but I hope that's a good enough draft.


tests/CMakeLists.txt
```diff
@@ -8,13 +9,13 @@ PROJECT(TESTS)
...
-ADD_TEST(afni:start ../../tests/xvfb-driver -- afni -com "OPEN_WINDOW axialimage; SAVE_JPEG axialimage test1; QUIT")
+#ADD_TEST(afni:start ../../tests/xvfb-driver -- afni -com "OPEN_WINDOW axialimage; SAVE_JPEG axialimage test1; QUIT")
 
 # TODO: figure out how to reference it correctly, or just copy to this
 # target tests/ build directory
 # Allow for 1 error and 2 failures reported -- due to the ones upon exit
-ADD_TEST(do.examples ../../tests/xvfb-driver -e 0 -f 0 -- @DO.examples -auto_test)
-SET_TESTS_PROPERTIES(do.examples PROPERTIES LABELS GLX)
+#ADD_TEST(do.examples ../../tests/xvfb-driver -e 0 -f 0 -- @DO.examples -auto_test)
+#SET_TESTS_PROPERTIES(do.examples PROPERTIES LABELS GLX)
```
yarikoptic: so tests pass since they are now disabled? ;-)

tharvik: The examples are, I was unable to find a way to make them works, but the `src`'s ones are working.

tharvik:
@yarikoptic, I would be better to have a separate branch, discuting a patch of ~1300 is kinda hard to do here.

tests/CMakeLists.txt
yarikoptic: well, if you leave them in so we could see what/how is failing may be we could help resolving those?
FWIW -- currently master is passing all the tests: https://travis-ci.org/afni/afni

yarikoptic:
@tharvik will do!  actually I had already somewhere a branch, where I was trying to expand our cmake build infrastructure to cover all "missing" aspects of upstream code.  But it was done on the elderly (year old?) version of afni.  I will buzz you whenever I get to that (hopefully later today/tomorrow)

tests/CMakeLists.txt
tharvik: I've failed something my side then, I'll try to fix it in the next days.

nno:
@tharvik thanks for the contribution, very impressive!
What are your thoughts on possibly waiting for @yarikoptic 's branch with improved neurodebian support before asking review by the AFNI developers (in particular @afni-rickr who does most github merges)?

tests/CMakeLists.txt
nno: If travis tests failed due to a conflicting ``count`` binary: please note that this was resolved a few days ago in (already merged) PR https://github.com/afni/afni/pull/58 after reporting issue https://github.com/afni/afni/issues/56.

yarikoptic:
@tharvik Here is our set of patches on top of elderly AFNI:  #61

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
-    # to gain access to built binaries, include path first
-    # (otherwise the wrong `count` binary may be used in the tests)
-    - export PATH=$PWD/$FLAVOR:$PATH
-    - mkdir ../testing; cd ../testing # from here will run tests
-    - cmake -DAFNI_BUILD_TESTS:BOOL=ON ..
-    - ctest --verbose
-    # Smoke tests for some additional commands
-    - 3dinfo
+    - mkdir build; cd build
+    - cmake ..
```
yarikoptic: tried to build with `BUILD_SHARED_LIBS` to seems run into some circular depends/includes...
```
 CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
   "mri" of type SHARED_LIBRARY
     depends on "rickr" (weak) 
     depends on "niml" (weak) 
   "niml" of type SHARED_LIBRARY
     depends on "mri" (weak)    
     depends on "rickr" (weak)
   "rickr" of type SHARED_LIBRARY
     depends on "niml" (weak)    
     depends on "mri" (weak) 
 At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.

```
would be nice to have a matrix runs within travis setup to build both ways

cmake/FindXp.cmake
```diff
@@ -0,0 +1,17 @@
+find_package(Xext)
+find_package(Xau)
+
+find_package(PkgConfig)
+pkg_check_modules(PC QUIET xp)
```
yarikoptic: I believe use of the Xp library was deprecated in #30 so FindXp/linking to libXp shouldn't be needed... I could be proven wrong though

afni-rickr: This depends on the OS.  In Debian and Ubuntu, libXp is not needed, but it is in Red Hat based systems (openmotif) depends on it.  I am not sure about Gentoo.

tharvik:
So, I tried to introduce some fixes with `BUILD_SHARED_LIBS`, but there seems to trigger many unseen bugs. I was thinking about taking the @yarikoptic patched branch and try to integrate my changes. It would provide a nice way to ensure thats building and testing are passing (as I guess it is the case now). WDYT?

yarikoptic:
Well... For our NeuroDebian branch then I might need to introduce more patches we had in the code as well... But it is just a few. And that worked only with a year old version of afni, so it wouldn't build current version. And it build primarily against system wide libraries - we did strip and sources of many copies it carried at that point.
On a positive side - we didn't have a dependency on a recent version of cake so it build fine across wide range of Debian releases.

I will prep a branch later on today. Cheers

cmake/FindXp.cmake
TheChymera: libXp has been deprecated on Gentoo.

PR #61, opened 2017-09-05
DO NOT MERGE: Minimalistic set of patches from NeuroDebian build introducing alternative CMake infrastructure

yarikoptic:
This is just a small excerpt from all the patches which we apply for neurodebian build

This one is based on an elderly AFNI_16.2.07 release -- the last one we updated cmake files for.

It is not complete to build the entire AFNI, and is just for demonstration purposes for the purpose of #60


yarikoptic:
too old and obsolete -- work in `master` by @leej3 is replacing it

PR #64, opened 2017-09-12
FIX: correct issue in AFNI Matlab library "unsplit_string" function

nno:
Minor fix in the AFNI Matlab library.

``unsplit_string`` produces incorrect results if more than two strings are "unsplit". This results in incorrect NIML files being written.

ping @afni-rickr @afni-gangc

### DIFF
--- a/src/matlab/unsplit_string.m
+++ b/src/matlab/unsplit_string.m
@@ -23,8 +23,8 @@ end
 parts=cell(1,2*n-1);
 for k=1:n
     parts{k*2-1}=c{k};
-    if k>1
-        parts{k}=sep;
+    if k<n
+        parts{k*2}=sep;
     end
 end
 


afni-rickr:
Thanks @nno! 
Matlab isn't my thing, but the solution looks like it will add an unneeded separator if n==1, if that matters.
What about: if k > 1: parts{2*k-2} = sep;

nno:
Hi @afni-rickr, thanks for taking a look.
Remember that Matlab uses one-based indexing. With ``n`` elements, the ``for k`` loop makes ``k`` have values from 1 up to and including ``n`` -- not from 0 up to and including (n-1) as in the C language.
So if ``n==1``, then the first loop iteration has also ``k==1``, the guard ``if k<n`` evaluates to false, and no separator is added.

See also Matlab output:
```
>> unsplit_string({'foo'},'XX')

ans =

foo

>> unsplit_string({'foo','bar'},'XX')

ans =

fooXXbar
```

That said, I think your solution would work equally well :-)

afni-rickr:
Oh, right!  Then your code reads better.

nno:
All right, thanks for merging.

PR #67, opened 2017-10-26
Added notes for building on RedHat/CentOS 6

giovtorres:
These instructions work for building AFNI on CentOS 6, starting with a base/minimal installation.


afni-rickr:
Great, thanks!

PR #68, opened 2017-10-29
Added dockerfiles for building from source

giovtorres:
These could be useful for testing source builds.


afni-rickr:
This is wonderful, thanks!

PR #69, opened 2017-11-03
Don't add absolute include-path to <>-includes

Doeme:
Don't add absolute include-path to <>-includes


afni-rickr:
Odd.  I suppose the author had some bad system configuration, and they forgot to undo that path.  Thanks for the update!

PR #70, opened 2017-11-15
fix for -despike_mask option in afni_proc.py?

dillonplunkett:
Hello! 

The `-despike_mask` option flag to afni_proc.py seems to have no effect. 

Investigating this issue is the first time I've looked at the afni source, so this proposed change might be totally off the mark, but I think it addresses the problem. `-despike_mask` works as expected on my machine after making it, at least.


afni-rickr:
That is right, the incorrect option name was applied in the code.
It seems that not many people care about that option... :)  Thanks!

PR #74, opened 2018-03-08
making error messages more helpful & clear

raamana:
I ran into an issue today when AFNI tools were not able to write Nifti files to disk, and the supposed "`NWAD ERROR`" wasn't helpful to me identify the issue clearly. 

```
** NWAD ERROR: wrote 0 of 491520 bytes of brick 1 of 96 to file++ 3dTshift: ...
...
** NWAD ERROR: wrote 8192 of 491520 bytes of brick 1 of 96 to file++ 3dTshift: ...
```

Although it could be deduced indirectly from "wrote only x of X bytes" message, this failure could be made more explicit to establish it as a "write failure" and nothing else.. Hence this PR.

Also I had to read the source code to figure out `NWAD` meant Nifti Write All Data, which is still a guess at this point :)

### DIFF
--- a/src/nifti/niftilib/nifti1_io.c
+++ b/src/nifti/niftilib/nifti1_io.c
@@ -5160,14 +5160,14 @@ int nifti_write_all_data(znzFile fp, nifti_image * nim,
 
    if( !NBL ){ /* just write one buffer and get out of here */
       if( nim->data == NULL ){
-         fprintf(stderr,"** NWAD: no image data to write\n");
+         fprintf(stderr,"** ERROR: NIfTI WRITE FAILURE: no image data to write\n");
          return -1;
       }
 
       ss = nifti_write_buffer(fp,nim->data,nim->nbyper * nim->nvox);
       if (ss < nim->nbyper * nim->nvox){
          fprintf(stderr,
-            "** ERROR: NWAD: wrote only %u of %u bytes to file\n",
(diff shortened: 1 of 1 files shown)


raamana:
cc @afni-rickr

afni-rickr:
Thank you for the comment @raamana.  But actually those new error messages seem less helpful to me.  The one thing that might be added to the original ones is mention of NIFTI.  They already mention  errors when writing.  Also, without NWAD, a coder would have a lot more trouble locating the placement of the error in the code (there are very many places for write errors).  It is meant to make the error messages possibly unique in the code.

raamana:
If there is a need to prefix the message with an unique code such as `ERROR: NWAD:`, we can certainly do that, in addition to making it easy for the user to follow what's going on.

would you like me to change the message to `ERROR: NWAD: NIfTI WRITE FAILURE` , adding the the NWAD keyword, which might be helpful for developers?

afni-rickr:
I made a more general commit that gets 'nifti' into the error messages that did not have it, and also added an 'only' for that exact case you mentioned (as in "wrote only...").  That seems more reasonable, so as to not put identical WRITE FAILURE text in so many places.
But you are editing nifti1_io.c, which is effectively no longer in use (by AFNI).  These changes are to nifti/nifti2/nifti2_io.c.  See https://github.com/afni/afni/commit/c37fb14e4e8a3ebe1d6bb26f27453719c6b5f976 .

afni-rickr:
Okay, thanks for your comments.

PR #117, opened 2019-10-03
Fix implicit function declarations

pn2200:
Fix all compiler warnings about implicit function declarations (warning: implicit declaration of function) in the default build configuration for Mac OS X 10.14 by including appropriate headers, using macros that were already defined for functions related to mcw_malloc, adding forward declarations, or by calling the correctly named function.

### DIFF
--- a/src/3DEdge/src/extrema.c
+++ b/src/3DEdge/src/extrema.c
@@ -20,6 +20,7 @@
  */
 
 #include <extrema.h>
+#include <convert.h>
 
 static int _VERBOSE_ = 0;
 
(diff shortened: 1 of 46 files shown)


afni-rickr:
Hello @pn2200.  Thanks for the commits!  Would it be possible to separate the first 30 commits again?  I am seeing how this fits with the current systems, and there are some issues in that second batch.  If the first batch works out (I am happy to try the older 30 commits as a single merge), then maybe we can work through the second batch in smaller pieces.  Does that seem reasonable?  Thanks!

pn2200:
Yes, there is a problem with 379cdd1, which I haven't been able to figure out. I will close this PR, and I have opened a new PR (https://github.com/afni/afni/pull/120) based off a new branch with up to 4e44fd8 (the first 30 commits).

PR #122, opened 2019-10-10
Fix implicit function declarations

pn2200:
Fix additional warnings about implicit declarations of functions on Ubuntu, so the current Travis CI build should not have any warnings about implicit declaration of functions.


pn2200:
I will actually undo the re-ordering, because all the prototypes should be in SUMA_display.h.

PR #123, opened 2019-10-18
add c changes accumulated during build development

leej3:
REPLACE_XT can now be used to swap out X11's memory allocation.
f2cdir  main.c does not need a main function because it is only
used as part of the f2cdir shared object library.


leej3:
comments addressed

PR #126, opened 2019-11-12
gui_fix

leej3:
Segfault when opening AFNI gui. Would be great if @yarikoptic  had a fix...

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,6 @@
 **/__pycache__
 tests/afni_ci_test_data
 tests/output_of_tests
+tests/sample_output_of_tests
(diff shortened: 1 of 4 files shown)


yarikoptic:
This PR has no actual test added!

leej3:
correct. now it does!

afni-rickr:
It might make sense to test afni using "afni -no_detach".  Testing while it forks a sub-process might confuse things...

yarikoptic:
> It might make sense to test afni using "afni -no_detach". Testing while it forks a sub-process might confuse things...

sounds good

PR #127, opened 2019-11-13
fix_suma_and_add_system_install_management

leej3:


leej3:
@yarikoptic ... segfault!!!

A good days work.

yarikoptic:
Woohoo - success we waited for! ;-)

leej3:
@yarikoptic, I may need your help with this.I thought the error was to do with GLW but I now think that is incorrect.  

 I have failed to replicate the error on two Debian systems now (9 and 11). With that said, I observe precisely the behavior you describe with an ubuntu docker container used for our development (and so, I observe it on CircleCI).

PR #133, opened 2019-11-21
move audio script to scripts directory

leej3:
@afni-rwcox can we move this script? Just to keep things consistent...


afni-rwcox:
This is still experimental and I'd prefer it remains in the Audio sub-directory.

leej3:
sounds good. i didn't realize that.

PR #140, opened 2019-12-16
Harmonize f2c with system installed version

leej3:


leej3:
@afni-rickr, this is now working. Let me know if there's anything you want to check

PR #144, opened 2019-12-31
Fix typo in macro _AFNI_SETUP_HEADER_ used as header guard

pn2200:
Using the CMake build of AFNI on macOS 10.15 with AppleClang, there are 107 instances of header-guard warning:

```
../src/afni_setup.h:7:9: warning: '_AFNI_SETUP_HEADER_' is used as a header guard here, followed by #define of a different macro [-Wheader-guard]
#ifndef _AFNI_SETUP_HEADER_
```

Correcting the defined macro eliminates these warnings.

### DIFF
--- a/src/afni_setup.h
+++ b/src/afni_setup.h
@@ -5,7 +5,7 @@
 ******************************************************************************/
 
 #ifndef _AFNI_SETUP_HEADER_
-#define _AFNI_SETUP_HEADER
+#define _AFNI_SETUP_HEADER_
 
 #include "mcw_malloc.h"
 


leej3:
Thanks for this. Just a point to note, I use homebrew to install llvm which has clang with omp support.

 I still have a few things to debug to have the cmake build working fully for OS X (when OpenGL dependencies are added) but that compiler will likely be the one supported moving forward.

pn2200:
Noted about which compiler will be supported in the future. I was just trying to make it clear that these warnings are visible locally, but not on the continuous integration builds on travis-ci or circle-ci.

afni-rickr:
Thanks!

PR #145, opened 2020-01-01
Replace deprecated functions

pn2200:
Replace several usages of deprecated functions. The deprecated function `drem` is replaced by the C99 standard function `remainderf`, the deprecated function `tmpnam` is replaced by the POSIX standard function `mktemp`, and the deprecated function `tempnam` is replaced by the function `mkstemps` which is not standard by is available on Linux and macOS platforms at least. These replacements resolve several compiler warnings regarding deprecated declarations.

### DIFF
--- a/src/XmHTML/lib/common/LZWStream.c
+++ b/src/XmHTML/lib/common/LZWStream.c
@@ -777,7 +777,8 @@ LZWStreamInit(LZWStream *lzw)
 	lzw->uncompressed = False;
 
 	/* temporary output file */
-	tmpnam(lzw->zName);
+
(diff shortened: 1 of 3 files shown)


mrneont:
Thanks, these are some useful things to think about in terms of updating some functions. However, these particular changes do introduce changes to how the code would run, and so would not be viable options as formulated.  From time to time we have thought about finding more in-place replacements for these functions.  But we have to make sure we can still build+run on a variety of systems, and so we need to get our testing structure increased before such updates can go in. Hopefully we will get to that relatively soon.

PR #148, opened 2020-01-10
WIP: Improve packaging for cmake build

leej3:


leej3:
@afni-dglen, could you give this a try when on debian when you have a chance. This now checks explicitly for the suma error:

```** ERROR: glwDrawingAreaWidgetClass is NULL
   This might be an error in GLwDrawA.h where the class is
   not referenced using 'extern'.  An alternative is to use
   the local build of libGLws.a.```

I've encountered this repeatedly because homebrew and debian both ship the library in a form that breaks our build (the patch in the neurodebian repository has not yet been upstreamed). A quick fix is using -DUSE_SYSTEM_GLW=OFF, or resort to the GLw header modification [hack](https://github.com/afni/afni/blob/b6a9f7a21c1f3231ff09efbd861f8975ad48e525/src/other_builds/OS_notes.linux_fedora_25_64.txt#L23) by modifying the file installed on the system. Hopefully the longer term fix is that we upstream the change but for now at least an error from the undefined symbol occurs at build time rather than runtime.

Regarding the plugin error you encountered on OSX. Send on the error when you have a chance.  I can't reproduce it.

PR #149, opened 2020-01-13
reconfigure dset_ijk code for JL

afni-rwcox:
Let's see how this pull request thing works.
If there are any problems, JL get the blame.
If it works well, RWC gets the credit.
That seems fair to me!


leej3:
excellent. thanks. you may have your credit

PR #151, opened 2020-01-14
deprecate imcat for 2dcat

### LINKED ISSUE #128: Unbundling imcat
I just nocticed that afni ships its own version of [imcat](https://github.com/stolk/imcat). On Gentoo we have a package for that software. So I will just make AFNI depend on that. However, I would also like to stop AFNI from trying to get/install that file itself. How can I best do this?

leej3:
Resolves #128

imcat will continue to be distributed for some time
along with the 2dcat binary (identical)


leej3:
@mrneont

PR #154, opened 2020-02-11
improve python packaging

### LINKED ISSUE #130: Distribute python code as a package
@satra, I'll work through this soon. Could you add any insight gleaned during your recent dive down the rabbit hole here: https://docs.google.com/document/d/1ilvS2y6w51wglz3grVydGu6NyIQXs5Um9E8wtdJkakY/edit?usp=sharing

Cheers

leej3:
Resolves #130 

Typical distribution will keep the executable files on the PATH (remaining in abin) but all importable python functionality will be moved into afni_python.  This will 

+ allow namespacing:

`from afni_python import afni_base`

instead of:

`import afni_base`

+ enable an optional package install into a specific python interpreter (pip install -e afni_python). This helps to isolate the installation (for example installing the python code into a Conda environment) and enables external code to easily use afni's python functionality.
+ Will facilitate an easier install down the line for those of the python world who wish to simply use `pip install afni_python` and not have to do everything else... some thought will have to be put into support for tcsh functionality, some work is required to distribute pre-compiled c libraries/binaries via wheels or the like. This approach will likely exclude R dependencies and the GUI functionality of AFNI.

EDIT:  Ideally it would have been `from afni import ...` but the current distribution of afni (with executables in bin) prevent this because "afni" already exists in that directory. afnipy chosen to confirm with many XXXpy packages in the scientific python ecosystem


src/python_scripts/afni_python/db_mod.py
```diff
@@ -11958,7 +11958,7 @@ def show_help_examples():
             Please see 'auto_warp.py -help' for more information.
             See also -tlrc_opts_at, -anat_uniform_method.
 
-        -tlrc_NL_warped_dsets ANAT WARP.1D NL_WARP: import auto_warp.py output
+        -tlrc_NL_warped_dsets ANAT WARP.1D NL_WARP: from afni_python import auto_warp.py output
```
leej3: need to fix this

leej3:
I still need to check whether entry points will work if "python" is actually python2.

leej3:
Another few things would still need to be fixed:
+ module_test_lib.py functionality is a little broken as is
+ scripts in ptaylor would need some fixes for importing
+ the situation for executables will need some more thought. There is some funny variable scoping in align_epi_anat.py for example that precludes simply wrapping the code captured by `if __name__ == '__main__'` in a main function and calling it using an entry point.

src/python_scripts/afni_python/db_mod.py
afni-rickr: You do not need to go through all of this.

mrneont:
>     * ptaylo

What scripts in ptaylor are you referring to?  The 'make'- or 'cmake'-related ones?  Pretty much everything else there is just a C prog.

leej3:
> > ```
> > scripts in ptaylor would need some fixes for importing
> > ```
> 
> 
> What scripts in ptaylor are you referring to?  The 'make'- or 'cmake'-related ones?  Pretty much everything else there is just a C prog.

The python files. They import afni_base/afni_utils directly. This pull request would alter the import pattern to `from afni_python import afni_base`. But yes, my understanding is that none of that is distributed.

mrneont:
> > > ```
> > > scripts in ptaylor would need some fixes for importing
> > > ```
> > 
> > 
> > What scripts in ptaylor are you referring to?  The 'make'- or 'cmake'-related ones?  Pretty much everything else there is just a C prog.
> 
> The python files. They import afni_base/afni_utils directly. This pull request would alter the import pattern to `from afni_python import afni_base`. But yes, my understanding is that none of that is distributed.

Oh, you mean the things in ptaylor/supplement?  Those are ignoreable, I think-- those are just sundry notes to myself that I wanted easily accessible.

leej3:
Ok sounds good. I won’t worry about them so.

.docker/cmake_build.dockerfile
```diff
@@ -15,9 +16,15 @@ RUN  cmake \
...
     -DUSE_OMP=ON \
+    -DUSE_PIP=ON \
     $AFNI_ROOT
 
 RUN /bin/bash -oc pipefail \
 ninja -v | tee -a verbose_build.log 2>&1
 
+# Install
+RUN DESTDIR=installed ninja install; \
+    cd /opt/src/afni/src/python_scripts && pip install .
```
effigies: Simpler:

```suggestion
    pip install /opt/src/afni/src/python_scripts
```

src/python_scripts/afni_python/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    "1dplot.py",
+]
+
+do_not_import = known_py2 + not_importable + other_problems
+
+__all__ = [
+    f.stem for f in Path(__file__).parent.glob("*.py") if f.name not in do_not_import
+]
+
+from afni_python import *
```
effigies: This file is a very weird hack. I think I get it, but I don't see an advantage to it over  leaving `__init__.py` empty and letting `import afni_python.<bad_import>` fail when someone tries it.

leej3: Agreed. I was still trying to figure out organization etc. It served as an effective way to determine all missing imports/unimportable files. I'll remove it once the details of any large reorganization are determined

src/python_scripts/setup.py
```diff
@@ -1,88 +1,77 @@
...
-    'uber_ttest.py=afni_python.uber_ttest:main',
-    'ui_xmat.py=afni_python.ui_xmat:main',
-    'unWarpEPI.py=afni_python.unWarpEPI:main',
-    'xmat_tool.py=afni_python.xmat_tool:main'
-    ]}
+# Executable scripts obtained from the make target list_py_scripts
+ENTRY_POINTS = {
+    "console_scripts": [
+        "1d_tool.py=afni_python.1d_tool:main",
+        "1dplot.py=afni_python.1dplot:main",
```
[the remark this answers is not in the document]
leej3: Thanks. I didn't know about the scripts keyword. That should allow us to separate importable vs. executables cleanly with minimal renaming if desired.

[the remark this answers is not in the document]
leej3: > # User-facing scripts, installed via entrypoints
Just wondering what the advantage of these would be over scripts. Is that just for when we want to import from a file that also serves as an executable?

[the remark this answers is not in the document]
effigies: Yes, I think the main advantage is not needing to separate scripts from the rest of your source. Presumably also makes it easier to check that your tests cover your CLI code, too.

leej3:
closes #156

leej3:
This installs fine with something like:
`pip install -e 'git+https://github.com/leej3/afni.git@change_python_installation_approach#egg=afnipy&subdirectory=src/python_scripts'`

I think it [needs to have the -e](https://stackoverflow.com/questions/13566200/how-can-i-install-from-a-git-subdirectory-with-pip#comment60104679_19516714) for development mode and the quotes are required to stop bash from doing funny things with the `&`

PR #155, opened 2020-03-10
Improved RetroTS.py

gianfrancof:
This addresses the issue raised [here](https://afni.nimh.nih.gov/afni/community/board/read.php?1,155524,155524#msg-155524) 

it fixes errors thrown when only one of either respiration data file or cardiac data file is passed to the RetroTS.py.
It also automatically figures out what to write in output file based on the argument passed in.


afni-rickr:
Thanks @gianfrancof!

PR #156, opened 2020-03-10
🐍 lib_afni1D — Python 2→3

shnizzedy:
Ref https://github.com/leej3/afni/pull/13#issuecomment-597312662

> * Update [`lib_afni1D.py`]
> (https://github.com/leej3/afni/blob/a9831f41b8f769a09b7599aa888612ef474e1338/src/python_scripts/afni_python/lib_afni1D.py) and its dependencies to Python 3
> * Update [`setup.py`](https://github.com/shnizzedy/afni/tree/8ccdbb81ca294038c20740fc2152f54595a94a44/src/python_scripts/setup.py)
>     * URL (from `git+https://github.com/leej3/template_making.git` to `git+https://github.com/afni/afni.git@master#egg=afni-python&subdirectory=src/python_scripts`)
>     * version (from `0.0.1` to `0.0.2`)

— https://github.com/leej3/afni/pull/13

### DIFF
--- a/src/python_scripts/afni_python/afni_base.py
+++ b/src/python_scripts/afni_python/afni_base.py
@@ -1,8 +1,13 @@
 #!/usr/bin/env python
 
-# python3 status: started
+# python3 status: compatible
 
-import os, sys, glob, operator, string, re, afni_base
+import glob
+import operator
+import os
+import re
+import string
+import sys
 
 valid_afni_views = ['+orig', '+acpc', '+tlrc']
 valid_new_views  = ['+orig', '+acpc', '+tlrc', '']
@@ -30,7 +35,7 @@ class afni_name(object):
       if view in valid_new_views: self.new_view(view)
       return
 
-   def p(self):   #Full path 
+   def p(self):   #Full path
       """show path only, no dataset name"""
       pp = "%s/" % os.path.abspath('./')  #full path at this location
       fn = pp.find(self.path)      #is path at end of abspath?
@@ -39,7 +44,7 @@ class afni_name(object):
       else:
          return "%s/" % os.path.abspath(self.path)
 
-   def realp(self):   #Full path following symbolic links 
+   def realp(self):   #Full path following symbolic links
       """show path only, no dataset name"""
       pp = "%s/" % os.path.realpath('./')  #full path at this location
       fn = str.find(pp,self.path)      #is path at end of abspath?
@@ -67,16 +72,16 @@ class afni_name(object):
                              self.rangesel)
 
       if sstuff == '': return sstuff
-    
+
       return "%s%s%s" % (self.selquote, sstuff, self.selquote)
-      
+
    def ppves(self, quotes=1):
       """show path, prefix, view, extension and all selectors
          (colsel, rowsel, nodesel, rangesel)
 
          this is identically ppve(sel=1), but maybe without quotes
       """
-      
+
       # if no selectors, do not incude quotes    7 Apr 2015 [rickr]
 
       # if no quotes, clear and reset internal selqute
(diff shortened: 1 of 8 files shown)


afni-rickr:
Hi @shnizzedy .  Thank you for the interest.  It would be preferable for such a pull request to contain only important changes, omitting things like removed spaces and style changes.  It looks like this is really a URL change in setup.py, is that correct?

leej3:
Oh in that case this likely shouldn't merged. The principal reason for #154  is to allow more advanced users to install the python code into an environment for isolation/tying the installation to a python interpreter. It's true it serves as a stepping stone for a standalone pip installable package but there's a lot more to it than that. Having a pip installable package will only be especially useful when we have incorporated the C binaries into a pip wheel. 

Agreed with Rick on the formatting changes. In general it would be nice to have the style more pythonic and conformant with pep8  etc. but fixing the large diff that it generates should typically done in a separate commit and is often not worth it due to the obfuscation of the git history.

leej3:
@shnizzedy was there a specific use case you had in mind?

shnizzedy:
> It looks like this is really a URL change in setup.py, is that correct?

The second of these two commits, 8ccdbb8, is just a URL update in setup.py and a version bump; the first commit, a9831f4, is an update to several scripts (lib_afni1D, afni_base, afni_util, lib_timing, lib_vars_object & module_test_lib) from Python 2 syntax to Python 3.

> was there a specific use case you had in mind?

We use lib_afni1D in some unit and regression testing for C-PAC (eg, https://github.com/FCP-INDI/C-PAC/blob/a6528ff9a2aea810fea6786a636dcf703ce2a943/dev/circleci_data/scripts/compare.py#L239) and are working to remove Python 2 dependencies from our package.

I didn't find a Python 3 version of lib_afni1d, so I
1. forked this repository
2. ran 2to3 on lib_afni1D and its internal dependencies
3. went through the automated changes and make adjustments
4. pip installed from my fork and tested my use case
5. updated the URL and build number in setup.py
6. opened a PR

The changes to spacing outside of actual updates to the Python syntax are automated changes from 2to3 and Atom. If it would be better for afni's flow, I can add a commit to revert the style changes and keep the substance.

leej3:
Ok. Thanks. It's good to know that others are trying to use our python code. It provides an extra data point in support of #154. I will try to get it merged today. In absence of that it is worth bearing in mind a [hack](https://github.com/afni/afni/blob/8e6dbf7aaad26964127d23b40238dc4288c88a31/tests/scripts/utils/misc.py#L38) I used for this. 

> If it would be better for afni's flow, I can add a commit to revert the style changes and keep the substance.

Yes please, and thanks for the contribution.

shnizzedy:
>In absence of that it is worth bearing in mind a [hack](https://github.com/afni/afni/blob/8e6dbf7aaad26964127d23b40238dc4288c88a31/tests/scripts/utils/misc.py#L38) I used for this.

Oh, cool. Thanks. That's a nice little hack.

shnizzedy:
@leej3 @afni-rickr 

33f35c9 now includes the Py 2→3 changes with minimal whitespace and style changes

leej3:
This still doesn't change anything other than removing `from __future__ import print_function`, tweaking install url, altering imports (will happen on #154 anyway), and adding `# python3 status: compatible`. Let me merge #154, then you can rebase, squash and merge this if you want. Until then  the import pattern would cause this to fail tests  if they had sufficient coverage (which unfortunately they do not). i.e. the current `import afni_base` can only be changed to `from afni_python import afni_base` on that other branch. And note, @afni-rickr  and I decided this should actually be `from afni import afni_base`. I'll ping you when #154 is merged.

shnizzedy:
It sounds like #154 will make this whole PR unnecessary :+1:

leej3:
should be good now @shnizzedy, thanks for the help and let me know if you have any issues.

PR #160, opened 2020-03-24
Remove doom, a little too close to home these days

Shotgunosine:


jbteves:
I had no idea AFNI was being prophetic as well as humorous.

mrneont:
Have commented out that startup tip, can hopefully uncomment in sunnier days...

PR #161, opened 2020-03-26
move ohtml to lib

afni-rickr:


mrneont:
Looks good;  worked on Bootcamp data.

PR #167, opened 2020-04-17
Retrots bids

Shotgunosine:
Just adding a some options to make running RetroTS on BIDS formatted physio files a little bit nicer. Also found a bug where slice_order could not be specified as "custom" as in the documentation, only as "Custom".


leej3:
Looks like a test data problem. Will sort it out. Cheers

leej3:
@afni-rickr  could you take a look at this please

afni-rickr:
It looks okay (and passes my tests), though is bit arithmetic really desired in the `# Handle file inputs` section?

leej3:
Nice spot, thanks. @Shotgunosine your R is starting to leak through into your daily life.

Issue #139, opened 2019-12-12
create a c library for suma datasets 


leej3:
This will help modularize the code a little more as both suma and AFNI could link against the library to implement functionality with suma niml file formats. Additionally it would be useful because:

+ Currently X headers are required for building libmri. Isolating the suma dataset functionality from libmri would make it clearer as to how plausible a full removal of X as a dependency would be.

+ Encapsulating suma dataset functionality would reduce the burden of development of this functionality. The encapsulation would additionally aid external software in implementing functionality for suma datasets

+ Finally, it will aid efforts in extending the [niml schema](https://github.com/nih-fmrif/codeconvergence/issues/2) beyond AFNI datasets to SUMA datasets.

The wisdom of @afni-rwcox  shall be required for this...

Currently, naively removing SUMA_OBJS as sources during the build of mri results in the following missing symbols in the library:

```
src/CMakeFiles/cs_objs.dir/afni_environ.c.o: In function `AFNI_prefilter_args':
/opt/afni/src/afni_environ.c:480: undefined reference to `sphinxize_prog_help'
/opt/afni/src/afni_environ.c:536: undefined reference to `SUMA_IcoNums'
/opt/afni/src/afni_environ.c:544: undefined reference to `SUMA_IcoNums'
src/CMakeFiles/misc_objs.dir/vol2surf.c.o: In function `v2s_write_outfile_NSD':
/opt/afni/src/vol2surf.c:2487: undefined reference to `SUMA_CreateDsetPointer'
/opt/afni/src/vol2surf.c:2492: undefined reference to `SUMA_AddDsetNelCol'
/opt/afni/src/vol2surf.c:2502: undefined reference to `SUMA_AddDsetNelCol'
/opt/afni/src/vol2surf.c:2509: undefined reference to `SUMA_AddNgrHist'
/opt/afni/src/vol2surf.c:2522: undefined reference to `SUMA_FindDsetAttributeElement'
/opt/afni/src/vol2surf.c:2546: undefined reference to `SUMA_WriteDset_ns'
/opt/afni/src/vol2surf.c:2550: undefined reference to `SUMA_FreeDset'
src/CMakeFiles/thd_objs.dir/thd_atlas.c.o: In function `niml_atlas_label_table_to_atlas_list':
/opt/afni/src/thd_atlas.c:342: undefined reference to `SUMA_strtol_vec'
src/CMakeFiles/thd_objs.dir/thd_atlas.c.o: In function `get_xform_chain':
[...]

PR #168, opened 2020-04-21
Suma dataset extraction

leej3:
Addresses the first part of issue #139


afni-rwcox:
Remove non-X11 includes from suma_suma.h into suma_objs.h and use the latter for AFNI codes, since they don't need X11 stuff.  All that's left of suma_suma.h is the X11 includes and then a suma_objs.h include to pick up the macros and prototypes.

PR #171, opened 2020-05-08
Fix RetroTS.py custom flag and add a test

Shotgunosine:


leej3:
@Shotgunosine, just to clarify, is this fix for the custom slices behavior just for bids or for both use cases of retrots? And can we now remove ` # Does nothing, unsure of it's purpose`?

Shotgunosine:
In my last PR I broke the custom slices behavior for all use cases of retrots. 

I didn't add the "does nothing" line, so I can't comment on it's appropriateness. I know that having that line wrong breaks things, but I don't understand the code well enough to explain what it does. If you wanted to change it to "Does something, unsure of it's purpose" I wouldn't object.

leej3:
Ha. Very well. Both seem equally obtuse. We'll leave as is. Thanks for the
fix.

PR #172, opened 2020-05-08
WIP:Update cmake installation details

leej3:


leej3:
Apart from general improvements in the cmake build and its management of system dependencies this pull request also:

+ Fixes a bug in which CircleCI cache misses occur for the dev_base image (introduced by #180). It turns out that the default working_directory has a default value that changes across executors. 
+ Uses xvfbwrapper, a python based tool for Xvfb used for tests
+ Alters the docker image entrypoints to improve compatibility with singularity

PR #173, opened 2020-05-08
change plugin search behavior

leej3:
@afni-rickr, @afni-rwcox comments welcome. I can add this to the make build too if desired.

THD_MAX_NAME is apparently too large for macOS to happily return a path. Not sure whether a more extensive work around is desired... I could just alternatively use the max path size variable that apple's functions specify.


afni-rwcox:
As long as it is compatible with Linux et cetera – code that is specific to Apple should go inside something like the below:

#ifdef DARWIN
  call Apple specific system functions
#endif

src/afni_plugin.c
```diff
@@ -305,7 +305,6 @@ ENTRY("PLUG_get_many_plugins") ;
    if( epath == NULL )
      epath = getenv("AFNI_PLUGIN_PATH") ; /* try another name? */
 
-#ifdef WHATS_MY_EXEPATH
```
leej3: @afni-rickr, removing this ifdef seems to cause a segfault. Any ideas why that would be?

PR #174, opened 2020-05-11
add FORCE target and use it for afni_src.tgz

leej3:
@afni-rickr, this sets the afni_src.tgz target to always be out of data. A pre-existing src/afni_src.tgz may have bitten me...


leej3:
closing. instead will make sure  tgz is removed by make cleanest

PR #186, opened 2020-07-17
Coxplot troubleshooting

leej3:


leej3:
fixes a disparity between make and Cmake build observed during distribution build. coxplot.h doppelgänger eradicated

Issue #194, opened 2020-07-30
/etc/localtime mounting error while running singularity


evansorenson:
Hello,

I am trying to run the afni dockerfile in singularity. I tried binding /etc/ in the command line and setting the TZ environmental variable beforehand but still no luck. I'm not really sure if this is an issue with the afni dockerfile or singularity itself, and I couldn't find a whole lot on this error. 

> FATAL:   container creation failed: mount /etc/localtime->/etc/localtime error: while mounting /etc/localtime: could not 
> mount /etc/localtime: not a directory

leej3:
Hi @evansorenson. Thanks for raising the issue. Could you post the singularity command you are executing?

leej3:
I see a warning  when I run a container:
```
singularity pull docker://afni/afni_cmake_build
singularity run afni_cmake_build_latest.sif
```
It says 
```
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```

But the command has executed successfully:
```echo $SINGULARITY_CONTAINER```

Tells me the container I am now in.

**Comment by leej3** [comment]
Ah, I get your reported error when executed on my cluster. I'll let you know a solution.

**Comment by evansorenson** [comment]
Yup, I was also running it on my cluster. Great, thank you!

**Comment by leej3** [comment]
A workaround for this issue (for older versions of the image, and other images that have written to /etc/localtime in this way) is to run
```
singularity run --containall --bind /data afni_cmake_build_latest.sif
```

The --containall fixes the issue (works for v3.5.3 of singularity but not 3.6.1). You will unfortunately lose all the default directory binding behavior that singularity does for you to try to make things easier. The --bind /data is  an example of how to restore access to parts of the host file system from within the container. Thanks to to the NIH hpc staff for that bit of singularity kung fu!

PR #195, opened 2020-07-30
do not create contained time

leej3:
Resolves https://github.com/afni/afni/issues/194

A singularity behavior change causes the docker image to now break on some
clusters after conversion to singularity. This change may be reverted in future
versions of singularity. But who needs the time in a container?

https://github.com/hpcng/singularity/issues/5465

EDIT: it's really just the work around that the change in singularity behavior fixes. Otherwise, the images are broken for usage with singularity on systems that bind in this way. Good to know... hopefully its fixed now


cclerget:
@leej3 Follow up in https://github.com/hpcng/singularity/issues/5465

PR #196, opened 2020-07-31
cmake: remove volpack, plug_render, and SurfLocalStat

leej3:
fails due to some missing symbols in libmri but should be close to working
without this dependency


leej3:
Ok, mri_render.c is the culprit as suggested by @afni-rickr. Silly me.

leej3:
superseded by #197

PR #197, opened 2020-07-31
remove netcdf and volpack dependencies

leej3:
I think this is good... the great arbiter circleci shall tell.

 @afni-rickr, @afni-rwcox could you take a quick look?


afni-rickr:
@leej3 That seems okay.  Though I am not sure we want to require setting DONT_ALLOW_MINC for the build to work.  Maybe we should reverse the logic so that the default is to not allow it:
#ifdef ALLOW_MINC_STORAGE ...

leej3:
Absolutely. Or even remove the code. Whatever you guys are happy with...

leej3:
This is what it would look like with a code removal.

I think the following two git aliases might be helpful for such a thing:

```git config --global alias.lc 'log --name-only  --pretty=format:"%n%n%nDeprecation: %s%nTimestamp: %ai%nCommit: %h%nAuthor: %an%nFiles affected:" HEAD~1..HEAD'```

```git config --global alias.deprecate '!git lc >> .deprecation.txt;git add .deprecation.txt; git commit -m "Code removal"'```

After changes to build/licence/docs etc. one can add all of the C code changes in a single commit. Then running "git deprecate" will log all of the files changed, who dunnit, timestamp etc. 

Subsequently if someone wished to restore the functionality, or just check how something was implemented they can do a quick search through .deprecation.txt

leej3:
Ah yes. It is. And I didn't remove the source file. Are we keeping the source for that? 

Option A: remove the source and make directives for it here
Option B: leave as is (you didn't want it distributed anyway).
Option C: Pull it out of this pull request and make it an optional target in the cmake build separately
Option D: Not option A,B, or C

afni-rickr:
B seems good.  This is something that we might want to eventually distribute, but not now.

leej3:
Cool. Sounds good. Merge at will.

PR #198, opened 2020-08-13
testing improvements

leej3:


leej3:
This pull requests tidies continuous integration for for the afni codebase


## General improvements for test execution
Rather than resorting to bash scripts etc, testing functionality has
been consolidated into a python script run_afni_tests.py. The scripts
functionality itself is heavily tested so hopefuly moving forward this
will provide a robust interface to testing the full codebase.

The tests now run reliably on MacOS (with known failures Xfailed as
required)

Parallel test execution now works well. Race conditions for test data
download, Xvfb buffers, communication ports, Xquartz displays, and
environment variables have all been fixed so the more cpus the faster
the execution of the test suite. pytest-xdist was essential for this
(over pytest-parallel) because of the isolation it provides between
running tests. Additionally, lockfiles are used for the other race
conditions (process locks that worked with pytest-parallel will not work
here).

Timeout functionality is added for testing. This can be very useful some
times for commands that are known to hang across different executable
versions. It is also required for failures of suma, which when
backgrounded and listening on a port will never close...


## Coverage testing
The coverage setup originally added by Ross Markello has been restored.
Previously, using coverage inflated the size of the development image to
18GB. A number of changes have been made this less of a problem. Among
others, the cmake build is now used, which uses shared libraries for all
libraries.
Additionally certain aspects of the build/CI infrastructure have altered
1. mrilib now has no X dependency
2. mcw_malloc memory stack monitoring is no longer used
3. the circleci setup is more modular and tag-specific jobs are being
 used to reduce baseline resource usage.

Running coverage every push to github seems excessive but updating the
coverage for every release is now manageable.


## Circleci/docker improvements
Circleci testing has been improved in a number of ways. The details of
data caching are now encapsulated in the python tool (permissions get
tricky as the user needs to be root on circleci but likely won't be
anywhere else). Additionally the executor used across the different jobs
has been homogenized.

Circleci executor changes were previously explored in #211. The solution
used here is to define an afni executor (docker image
afni/afni_circleci_executor) that contains all dependencies required on
circleci. The advantage of this is that now all circleci jobs can be run
locally using the circleci local CLI. The downside is that caching
test-data becomes more convoluted.

circleci mounts the host docker service into the executor as described here:
http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/

The circleci executor now has docker v19.03, which supports chown upon
copy. This chown on copy drops the size of the image. It also gives the
make_build access to /opt/afni/src (previously it was not worth the
increased image size to chown it)

run_afni_tests.py now manages test data during circleci testing. As
suggested by the circleci docs, it creates a container and mounts a
volume to it and then shares that volume with the main testing container
using the "volumes-from" option. Trickiness regarding
permissions/ownership are handled along with this.


## Testing graphical user interfaces
The gui testing has proven extremely different to get working. The main
challenges were working with Xvfb (xvfbwrapper provides a more pleasant
interface from a pythonista point of view), race conditions for ports,
and the fact that suma does not raise non zero exit codes and hangs
indefinitely when a failure does occur. Additional challenge was working
out how to get opengl functionality working on MacOS using Xvfb (I
believe it is not possible unless Xquartz is recompiled with indirect
GLX support for Xvfb frame buffers). The solution was to run all suma
tests on the physical display of MacOS with the appropriate locking to
prevent race conditions.

## Test data management
Test data for the testing suite is handled using datalad. This allows
convenient management of recursive git-annex repositories so large data
trees can be represented on disk without all of the data actually data
being there. Data management should not have afnipy as a hard
dependency. As tests are executed data that only exists as a symlink is
downloaded. Race conditions previously stymied robust behavior here but
it is now working really nicely. It has been encapsulated in the
data_management.py module for ease of reuse.

PR #206, opened 2020-09-15
more backups for linux_distribution

afni-rickr:


leej3:
superseded by #206 and #208

PR #211, opened 2020-10-02
update circleci executors

leej3:


leej3:
@mrneont this updates the circleci executors used. Updates on 2020/10/26 will remove some of the current circleci virtual machines (well actually 'descale', but its a slippery slope from there!). Additionally older versions of  remote docker will be removed.

For information on remote docker on circleci see this:

https://circleci.com/docs/2.0/building-docker-images/

A more recent version of remote docker is now explicitly defined to avoid any issues when the deprecation does occur.

Previously for the deployment step (pushing docker images to docker hub upon version updates) the virtual machine circleci/classic:201711-01 was used. This has been replaced here with the docker image docker:18.01.0-ce-git along with remote docker. This necessitates some fixes because the docker image is far more minimal (alpine OS is used). The upside of all of this is that the deployment step can now be executed easily on any machine that has docker and circleci installed with something along the lines of:

    circleci local execute --job=deploy_cmake_build

leej3:
I will say that this is not entirely complete though. I have not changed the executor for the tests step. And I want to explore the most efficient way to execute with respect to the test data. I can't directly mount previously cached test data into the container when using remote docker. Option 
a) Do a more convoluted usage pattern for mounting the data
b) Forget about caching the data.

I'm leaning towards a). Less needless data downloading.

PR #212, opened 2020-10-06
AFNI's Python real-time interface updates

roopchansinghv:
This is a larger pull request, to address the socket timeout issue, and also to redirect code to repository where the real-time interface is being enhanced.


leej3:
Let me know if you want help fixing the build on circleci...

roopchansinghv:
Not sure why this failing.

One commit added one line of python code to an existing library file.

The other commits removed some code that's not used anywhere else in AFNI, and replaced it with a reference to a repository I am working on that contains the identical (but actively supported) code.

Is this failing b/c I am not logged into CircleCI?

leej3:
No, I try to handle all of the python code along with the build. So there's
a few lines of code in the CMakeLists.txt files that needs to be removed.
Around here:

https://github.com/afni/afni/blob/b5eca3313db4ab4992e80502d0b760893c0d7fd9/src/python_scripts/CMakeLists.txt#L81

roopchansinghv:
Awww - why did you have to be so thorough? ;-)

Okay - it seems most of the tests are passing.  Let me see if I can resolve the last issue.

leej3:
Nice. You passed the cmake gauntlet. Now for the make build!

roopchansinghv:
So the logs are complaining it can't find the files I removed (i.e. the 'demoExpt.py' file and 'afniRTI' directory).  I thought removing those entries from the CMakeLists.txt you referred to would fix that.

Is that cached somewhere else?  I'll check through the tree a bit more.

leej3:
You did fix it yes. This is different error in the build that uses Makefile.Include (as in the older build system that we are migrating from).

So this file:
https://github.com/afni/afni/blob/d02f5474b2e4a833735e4b0e7a9eea266d5e127a/src/Makefile.INCLUDE#L3694

roopchansinghv:
Okay - just cleaned up Makefile.INCLUDE

Hopefully integration tests will get automagically kicked off again.

leej3:
They should kick off but haven't for some reason. And it's not showing up. We can just wait or you can make some change and push that to wake it back up :) A useful tweak you could make to kill two birds with one stone  would be to rebase and force push. Something along the lines of:

```
git remote add afni https://github.com/afni/afni.git
git fetch afni
git rebase afni/master
# You need to force push as the history will have changed. 
# And you may have to replace roopchansinghv with origin
# This assumes you are using your master branch locally
git push roopchansinghv -f master:master
```

Or merge afni/master into your branch and push that. Whatever your preference is...

roopchansinghv:
Hmmm - okay all commits are now showing up (after I tried your suggestions above - thanks!) - but I don't see the integration tests running.

leej3:
> but I don't see the integration tests running 

Neither do I. I haven't seen this before. 

I think it might be that we have run out of circleci credits. We shall look into that.

All the tests pass when I run them locally though so I think this is all good from the perspective of CI.

@afni-rickr  could you take a quick look at this?

afni-rickr:
To be sure, @roopchansinghv, you want to leave your directory in afni_src.tgz for README.md, is that right?  This looks good.  Though it seems the default timeout should be None, so is this getting altered under psychopy?  Just curious.

roopchansinghv:
Yes - wanted to just leave the README behind, to allow people to follow this, if they had been looking at the before.

As to the timeout - possibly.  I don't remember this timeout issue when I run the AFNI "demo 6" neurofeedback experiment with the default packages.  However, when I tested my version with the PsychoPy interface - the connection times out in approximately 20s if no data is sent to it.  I verified this again yesterday (2020-10-06), before I made these commits.

PR #218, opened 2020-11-01
Add some tests

leej3:


leej3:
I'm going to merge the additions to the testing functionality separately. this is now rebased onto master so includes fixes for OMP. That may remove the discrepancy between the output of the make/cmake build systems. If it doesn't then some debugging is going to be required to figure out why it's happening.

To make the test data easier to modify i am going to host it on g-node (https://gin.g-node.org/leej3/afni_ci_test_data). That will remove issues with pushing when not on the VPN

PR #227, opened 2020-12-01
deindent RetroTS.py slice_order if statement

pab2163:
fixing if statement for reading in custom slice acquisition timing files (e.g. for multiband data)


afni-rickr:
Thanks!

PR #228, opened 2020-12-01
fix for 622e127eea57c5

leej3:
RetroTS.py needs to be executable for testing.


leej3:
@afni-rickr or @Shotgunosine, let me know if you spot a fix for the remaining failures in RetroTS.py. Thanks!

PR #231, opened 2020-12-04
Resampling

mrneont:


mrneont:
This is the start of reporting more about the ijk_to_xyz matrices in a dset--- what we refer to presently as aform_real, aform_orth and aform_card.  

Presently, we are just using this information in 3dinfo--hence, only reporting on it, not *doing* anything with it yet.  Knowledge is half the battle!

Read the "Options for displaying ijk_to_xyz matrices" note about the matrices (in general, and specifically the aform_* ones) and the new opts for reporting on them.

PR #234, opened 2020-12-17
add mri_mini as a mechanism for encapsulating some of afni's functionality for a web assembly build.

leej3:
@afni-rickr if you have some time over the next few days could you have a look at this. All the circleCI tests are passing so I think there should be no issue with the build used for distribution but perhaps you could confirm...

@afni-rwcox  is pondering a more substantial attempt at modularization/spring cleaning of the libmri headers but it may be that merging this as a separate set of changes beforehand makes that process more managable.


afni-rickr:
@leej3 
This seems okay.  Just as a test, I am building a test program (test_mri_mini) that uses such a library (just via the make system).  Should I add it to this branch?

leej3:
Yes please. That would be great.

afni-rickr:
@leej3 
Added.  It assumes the shared library method that is not yet available on macs.

afni-rickr:
@leej3
Is that `brew --cask` error one that needs fixing now?

leej3:
Yes, on master or here. I’ll take a quick look. I don’t think it will be too bad.

leej3:
I'll merge this branch for now. I'll update with various improvements later

PR #241, opened 2021-01-05
WIP:Add some tweaks to the cmake build for gentoo packaging

leej3:
To begin, the cmake variables used to search for the python interpreter at configure time are user modifiable.

Another that detail that we need to solve is python code installation. Ideally this will use the portage distutils-r1 eclass mentioned by @TheChymera. If this works out then COMP_PYTHON should probably be set to off.

### DIFF
--- a/.docker/afni_dev_base.dockerfile
+++ b/.docker/afni_dev_base.dockerfile
@@ -66,7 +66,6 @@ RUN apt-get update && apt-get install -y eatmydata && \
     freeglut3-dev \
     git \
     libf2c2-dev \
-    libglew-dev \
     libglib2.0-dev \
     libglu1-mesa-dev \
     libglw1-mesa-dev \
--- a/.docker/cmake_build.dockerfile
+++ b/.docker/cmake_build.dockerfile
@@ -20,7 +20,7 @@ RUN \
     fi; \
     cmake \
         -GNinja \
-        -DCOMP_BINARIES=ON \
+        -DCOMP_COREBINARIES=ON \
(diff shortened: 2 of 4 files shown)


leej3:
It looks like providing the python source subdirectory, `src/python_scripts`, to the esetup.py function (see the portage [eclass description for distutils-r1](https://devmanual.gentoo.org/eclass-reference/distutils-r1.eclass/index.html) should be sufficient. This should drive the appropriate installation of python source code. And to confirm, COMP_PYTHON should be set to OFF.

The current cmake build system either copies python files into the binary directory to mimic the current packaging approach (see
[here](https://github.com/afni/afni/blob/162e164e55393ee65733a3447b19550b190f3cda/src/python_scripts/CMakeLists.txt#L59)). Alternatively, by default, it installs the afnipy package in development mode (source files are left in place and egg-info is add to the site packages of the "current" interpreter). This latter step is done [here](https://github.com/afni/afni/blob/162e164e55393ee65733a3447b19550b190f3cda/cmake/make_sure_afnipy_is_installed.sh#L35).

@afni-rickr,  I remember you were asking about this... adding `-DSTANDARD_PYTHON_INSTALL=OFF` should make the python code fully contained in the path you specify with DESTDIR.

afni-rickr:
@leej3 Sorry for being slow.  Is this ready to be merged?  Thanks!

leej3:
Not a problem. I’ve marked it as work in progress now. I’ll take a quick look with the @TheChymera this week to see if we can finish it off.

PR #250, opened 2021-02-10
Drg mac1015

afni-dglen:


afni-rickr (approved):
Perfect.  I have never seen such good code.

PR #260, opened 2021-04-07
Fixes/3d ecm

ccraddock:
Fixes to improve 3dECM:
- updates sparse_array.c to abandon building similarity matrix as soon as memory budget is exceeded (previously would try to add a few more connections which would result in a string of out-of-memory errors)
- fixed ridiculous seg fault that would occur when exiting on a error
- changed it to only output binary ECM result when the -do_binary flag is specified, rather than every time
- added descriptive label to the output buckets.

### DIFF
--- a/src/3dECM.c
+++ b/src/3dECM.c
@@ -160,10 +160,8 @@ static long total_mem = 0;
 \
         if( eigen_vec != NULL ) \
         { \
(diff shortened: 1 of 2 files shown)


mrneont:
These changes seem to only affect 3dECM (and sparse_array, which only gets used by 3dECM).

PR #263, opened 2021-04-26
Fixes/3d ecm

ccraddock:
- found a problem with 3dECM where the automatic threshold optimization when given a sparsity target would use a ton of memory. This is an issue when there are a bunch of time series for which the correlation is zero. Avoided this by eliminating all zero variance voxels prior to calculation.
- updated 3dLFCD to be more in sync with the original publication by changing its default neighborhood to the six face touching voxels. The user can change this to faces+edges or face+edges+corners (the prior default) using command line arguments. 
- improved documentation for 3dECM and 3dLFCD


mrneont:
Seems good:  all checks passed and no conflicts with base branch.

PR #269, opened 2021-05-09
Fix 3dBrickStat nan option text

### LINKED ISSUE #267: Option "-nan" reversed in 3dBrickStat help page
The 3dBrickStat help page reads:
<pre>
  -nan = include only voxel values that are finite numbers, 
         not NaN, or inf. -nan forces -slow mode.
  -nonan =exclude voxel values that are not numbers
</pre>

These two options sound like they both do the same thing: filter so only finite numbers are included. However, it seems that -nan actually only includes NaN/inf numbers (non finite or non numbers). I don't have a data set with NaNs to fully test, but on the data sets I've used `3dBrickStat -count -nan <file>` returns `0` while `3dBrickStat -count -nonan <file>`  returns the same as `3dBrickStat -count <file>`.

YurBoiRene:
-nan option in 3dBrickStat stated that it only included finite numbers but it actually omits finite number. This pull request fixes that.

Closes #267


afni-rickr:
Thanks for the suggestion, @YurBoiRene.  I will merge and then make additional clarifications.

Issue #278, opened 2021-06-12
thd_incorrelate and OMP


leej3:
I am getting an error when compiling thd_incorrelate.c in the cmake build on MacOS.

```
FAILED: src/CMakeFiles/thd_objs.dir/thd_incorrelate.c.o
/Library/Developer/CommandLineTools/usr/bin/cc -DDYNAMIC_suffix=\".dylib\" -DREPLACE_XT -I../src/3DEdge/src -I../src/niml -I_deps/nifti_clib-build/nifti2 -I_deps/nifti_clib-src/nifti2 -I_deps/nifti_clib-build/znzlib -I_deps/nifti_clib-src/znzlib -I_deps/nifti_clib-build/nifticdf -I_deps/nifti_clib-src/nifticdf -I_deps/nifti_clib-build/niftilib -I_deps/nifti_clib-src/niftilib -I../src/f2c -Isrc -I../src -I../src/rickr -DUSE_TRACING -DHAVE_ZLIB -DREAD_WRITE_64 -DLINUX -DDARWIN -O3 -DNDEBUG -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk -mmacosx-version-min=11.2 -fPIC -MD -MT src/CMakeFiles/thd_objs.dir/thd_incorrelate.c.o -MF src/CMakeFiles/thd_objs.dir/thd_incorrelate.c.o.d -o src/CMakeFiles/thd_objs.dir/thd_incorrelate.c.o -c ../src/thd_incorrelate.c
../src/thd_incorrelate.c:948:15: error: implicit declaration of function 'omp_get_thread_num' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
       ithr = omp_get_thread_num() ;
              ^
../src/thd_incorrelate.c:988:15: error: implicit declaration of function 'omp_get_thread_num' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
       ithr = omp_get_thread_num() ;
              ^
2 errors generated.
```

I can't remember too well but I don't think I ever managed to compile libmri openmp... without USE_OMP defined for it any instances of OpenMP code are appropriately removed by the C preprocessor... it seems this is not the case for thd_correlate.c

[...]

PR #268, opened 2021-05-09
Change CI test data to repo hosted on GIN

YurBoiRene:
Changes the CI test data repository [leej3/afni_ci_test_data](https://gin.g-node.org/leej3/afni_ci_test_data) hosted at [gin.g-node.org](gin.g-node.org). The repository on GIN does not require access to NIH or afni servers which makes it easier and more accessible to update test data. in the future.

Additionally, this repository is already updated with 3dClusterize data to fix the currently failing tests (Issue #266).

PR #279, opened 2021-06-12
Make some ci fixes

leej3:
This tries to address the issues in publishing test data with datalad described in https://github.com/afni/afni_ci_test_data/issues/2. 

It also includes a breakage in tests that was fixed in https://github.com/afni/afni/pull/268 but it does not look like that is going to be merged.

A temporary fix for 3dClusterize tests where the output diffing does not occur is added for now to pass the tests until the test data is updated appropriately.

There is still an issue with the compilation of libmri on MacOS described here:
https://github.com/afni/afni/issues/278

Additionally suma is no longer compiling on MacOS... that error looks more extensive.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -558,6 +558,6 @@ workflows:
               test_data_cache_version: ["5"]
           filters:
             tags:
-              only: /AFNI_\d\d.\d.\d\d/
+              ignore: /.*/
             branches:
-              only: /.*/
+              ignore: /.*/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,23 @@ include(get_git_repo_version)
 include(get_build_macros_and_functions)
 include(CMakeDependentOption)
 
+if( (NOT DEFINED ENV{CC}) )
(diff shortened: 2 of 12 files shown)


afni-rickr:
@leej3, @mrneont That CircleCI change seems good, though let's leave comments about the security issue.  @leej3, we can add those later if you prefer.

leej3:
Aha, thanks for that @afni-rickr.

afni-rickr:
It is nice to see the tests pass again, thanks!

mrneont:
Re: "Additionally suma is no longer compiling on MacOS... that error looks more extensive."  Does this refer to a CircleCi build error for Mac OS?  

If that is the case, I think that is resolved---there was two issues that I could tell, each just involving a tweak of "#include .." statements:  in one case, for SUMA_input.c, an "#include .." pointed to a package that was just on a local machine (but we that dependency wasn't necessary, so could be fixed just by commenting that line); in another case, for 3dNwarpCalc.c, the recent addition of an if-condition required putting two simple "#include .."s in one branch of the condition.  

@leej3 , could you please verify if these have resolved the issue(s) you were referring to?

Thanks.

PR #289, opened 2021-06-25
Int cmap range

mrneont:
Up until now, an INT_CMAP with N colors applies to an ROI dset with max value N, and sets its pbar max to also be N.

What *should* happen (and does happen now), is that an INT_CMAP with N colors applies to an ROI dset with max value N-1, and sets its pbar max to also be N-1.


mrneont:
Thanks for helping with this, RWC!

PR #291, opened 2021-06-26
improve omp behavior in cmake build

leej3:
addresses issues raised in #278

Now if USE_OMP is set by the user to OFF, an attempt at finding openmp is not done. In general
if OpenMP::OpenMP_C is not a defined target then afni targets do not link against omp or
define USE_OMP as a compile definition. Any deviation is a problem with the targets and should
be fixed. An example of a succesful implementation is:
https://github.com/afni/afni/blob/7bc375964a0a54323447842d6564a536e045476e/src/CMakeLists_binaries.txt#L250-L263


mrneont (commented):
This seems fine to me, and it is good that the CircleCI testing passes.

*But* note that I am not well versed in the ways of Cmake.

leej3:
Sure. I was just showing the logic of targets that can avail of omp. A few of your tools could make use of this (or perhaps they already do). 

Thanks for merging.

PR #294, opened 2021-07-21
Adds encoding if needed for abids_tool.py

### LINKED ISSUE #293: abids_tool.py fails on newer python versions
In Python 3.6, `subprocess` began returning byte arrays instead of encoding the bytes as strings. `b"x" == "x"` will evaluate to `False`, so on newer systems, `abids_tool.py` will crash because it believe that there is not a shape match between the string-encoded JSON and the byte array-encoded captured output. The captured output should be captured as ASCII.

jbteves:
Closes #293


afni-rickr:
Thanks @jbteves !

PR #305, opened 2021-09-28
Added support for IBM POWER8/9 architectures

pantaray:
Hello AFNI team!

We're running IBM POWER E880 servers in our HPC cluster infrastructure and I built AFNI for that architecture. Aside from putting together a corresponding Makefile, I had to tweak `gts` a little to make it work with `gcc-8` (I updated `config.guess` and hacked `ltmain.sh`). Maybe these changes are useful for others who want to build AFNI on similar hardware. 

All the best from Frankfurt, 
Stefan


neurolabusc:
@pantaray would you mind running the [AFNI Bigger Speed Test](https://sscc.nimh.nih.gov/afni/doc/misc/afni_speed/index_html) on your hardware? It would be interesting to see how this compares to modern computers from [AMD, Apple, and Intel](https://github.com/neurolabusc/AppleSiliconForNeuroimaging). I serve on my institutions research computing committee, and it is helpful to know about alternative HPC vendors.

afni-rickr:
Hi @pantaray.  This looks great.  Are the config.guess updates all from you, or have there also been updates to it upstream?  Just curious.  Thanks!

pantaray:
Hi @afni-rickr ! The `config.guess` changes are all upstream (taken from `automake-1.16`) in the hope to not unintentionally break anything on other platforms.

pantaray:
Hi @neurolabusc ! I tried to run the speed test but it seems to use a specific csh dialect and runs `3dDeconvolve_f` which I did not build (from what I see, `COMP_GUI` which triggers the build of `3dDeconvolve_f` is `OFF` in the default settings, only `COMP_PLUGINS` is `ON`). Is there maybe another AFNI benchmark available?

PR #312, opened 2021-11-22
Adds 3dDiff

jbteves:
Adds a program to compute the number of different voxels between two dsets.
(This is something I've found myself doing often without wanting to run 3dcalc and save to disk).
TODO:
- [x] default: short output summary with
- [x] 1-3D case: "Images do NOT differ" or "Images differ: n of v elements are different (p%)"
- [x] 4-D case: "Images do NOT differ" or "Images differ: n of v volumes contain differing elements (p%)"
- [x] `-brutalist`: output summary line only, 1-3D case shows [differs, n_differing_elements, total_elements, percentage], 4D case shows [differs, n_differing_volumes, total_volumes, percentage]
- [x] `-tabular`: Output table with summary line from `-brutalist` (Again, 1-3D case shows elements, 4D case shows volumes) and a table indicating for each row [differs, n_differing,


jbteves:
This is ready for review now, for some reason I can't request anyone to review it, though.

neurolabusc:
@jbteves you can also co-opt the code from the [niimath](https://github.com/rordenlab/niimath
) `compare` function. In my experience, it is useful to have a tool that can help detect both identical as well as equivalent results. For example, an x86-64 computer will provide different results for a 32-bit float [multiply-accumulate](https://en.wikipedia.org/wiki/Multiply–accumulate_operation) computation if you use the FPU (80-bit internal precision), SSE, or a specialized AVX FMA instruction (one less rounding). Likewise, there are often differences when one uses a different architecture (NVidia Cuda, aarch64, x86-64). For this reason, I find it useful to report the most extreme difference between voxel pairs. Consider comparing two 3D images where the r is identical (1.0, at least to machine precision): 

```
Images Differ: Correlation r = 1, identical voxels 73%
 Most different voxel -69.3133 vs -69.3133 (difference 1.52588e-05)
 Most different voxel locatoin 43x17x49 volume 39
Image 1 Descriptives
 Range: -472.393..491.385 Mean -0.00121971 StDev 6.8898
Image 2 Descriptives
 Range: -472.393..491.385 Mean -0.00121971 StDev 6.8898
```

jbteves:
Thanks for the tip @neurolabusc !

My intention here is to slowly expand the tool if people are interested. This is pretty much the "minimum viable product" for my use case, but there are significant other ways to expand the usefulness:
- Add code to match the data types, or else upcast the less precise one and then compare (this is much more tricky, and for my immediate purposes just coercing to float32 actually works pretty well).
- Add features to diff header elements. For example, if one file contains a header extension and another does not.
- Print out what the differing dimensions are.
- Intelligently determining whether the voxels diverge because they have different values, or if there is an implied mask (left has a 0.00--unlikely in floating-point for a real value, but right has a non-zero value; most likely this means that masks used to generate the images diverge, not necessarily that the underlying computations are different), and counting that up.
- (As you mention) the minimum and maximum divergence in values; to be truly fancy, fill out a histogram.
- Add Python bindings so that you can easily run diffs for testing frameworks such as pytest (any python implementation I've tried is just unbearably slow).
- Add a `-q` flag to match the `diff` behavior of the same name: Just stop if there are any differences at all and return non-zero.
- Add the ability to diff entire directories, which would let you do a little bit more seamless estimation of how much that "small change" impacted your data processing.

I'm sure there are other ways to improve things, too, but this is what I have off the top of my head. I do want to keep the program as light as possible; right now the runtimes are something like 250MB/s of diffing and I'm happy with that.

jbteves:
Team AFNI: I'm not sure what's up with `test_cmake_build`, but it doesn't look like this patch is the source. Let me know if I'm wrong and I'll try to fix it.

mrneont:
> Team AFNI: I'm not sure what's up with `test_cmake_build`, but it doesn't look like this patch is the source. Let me know if I'm wrong and I'll try to fix it.

@jbteves :  Indeed, it isn't your program, it's this currently open issue about the testing framework:
https://github.com/afni/afni_ci_test_data/issues/2
... that we haven't been able to update the "answer" data to reflect current code since a one particular update.

We are working with datalad folks to try to resolve this...

--pt

jbteves:
Is there a good way to parse scientific notation in C/AFNI that I'm not aware of? I feel like that would be a big QOL boost for this feature, but I'm not really excited about trying to roll my own in C.

mrneont:
> Is there a good way to parse scientific notation in C/AFNI that I'm not aware of? I feel like that would be a big QOL boost for this feature, but I'm not really excited about trying to roll my own in C.

What do you mean about "parsing"? There are format specifiers in C, for reading and writing scientific notation.  `%e` and `%g` are useful ones for scientific notation:

Consider this C program, for displaying scientific notation:
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

int main( void){
   float a = 100;
   float b = 1000000;

   printf("a as e:%e\n", a);   
   printf("a as f:%f\n", a);   
   printf("a as g:%g\n", a);   

   printf("b as e:%e\n", b);   
   printf("b as f:%f\n", b);   
   printf("b as g:%g\n", b);   

   return 0;
}
```
... which outputs:
```

a as e:1.000000e+02
a as f:100.000000
a as g:100
b as e:1.000000e+06
b as f:1000000.000000
b as g:1e+06
```
Note the difference is what 'g' does, depending on the value of the number.

Here is an example of reading in a text file of numbers, "file.txt":
```
1e5
2e-6
3e7
```

... and displaying them:
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

int main( void){
   FILE *fff = fopen("file.txt", "r");
   float a, b, c;

   fscanf(fff, "%e", &a);
   fscanf(fff, "%e", &b);
   fscanf(fff, "%f", &c);

   printf("a as e:%e\n", a);   
   printf("a as f:%f\n", a);   
   printf("a as g:%g\n", a);   

   printf("b as e:%e\n", b);   
   printf("b as f:%f\n", b);   
   printf("b as g:%g\n", b);   

   printf("c as e:%e\n", c);   
   printf("c as f:%f\n", c);   
   printf("c as g:%g\n", c);   

   fclose(fff);

   return 0;
}
```
Note how even reading as `%f` worked as expected here.  

Depending on the size of incoming number, one has to be careful about float/double/etc.

jbteves:
By "parsing" here I meant, "converting a series of character tokens into a floating-point number."
Wow, for some reason I thought `atof` would not parse scientific notation. Thanks for the pointers on using `fscanf` and the various format specifiers. I'll see if I can adapt the program to use that so that I can crash if the user specifies, say, "cat," as their tolerance.

jbteves:
Okay, another question: what is the preferred method of raising a user warning in AFNI? INFO_message or something else? I'd like to raise a warning if the user passes a negative tolerance since I have to sign-flip it.

mrneont:
Some communication options in AFNI C code:
```
INFO_message("What's up, Doc?");
WARNING_message("Danger, Will Robinson.");
ERROR_message("I'm leaving now.");
ERROR_exit("No. Goodbye.  I mean it.");
```

Separate note:  Something I haven't had time to do here with forking your branch, but I think you need to add "3dDiff" to one of the PROGS_? lists in Makefile.INCLUDE;  I don't think it matters which, but I put it into PROGS_4 to be able to have 3dDiff built on my Ubuntu 20.04.

jbteves:
@mrneont does this last commit mirror what you did locally?

mrneont:
@jbteves : yes.

jbteves:
Any changes you'd like to see on this?

mrneont:
@jbteves - I had some thoughts for the code, esp. on reading in the datasets initially.  I think that reading them in as THD_3dim_dataset types would be preferable to doing so as MRI_IMAGE types.  This should be more general, and allow for more standard flexibility of input naming, as well as verifying the consistency of multiple input dsets like for matching grids, etc.

In fact, this has inspired an update to 3dToyProg, or at least a separate kind of educational "starter" program with comments, with many inputs from @afni-rickr.  This is now in the code base: 3dEdu_01_scale.c (https://github.com/afni/afni/blob/master/src/3dEdu_01_scale.c).  I think this might be useful for tweaking the way inputs are done and reading in the data.  The output should be even simpler in your case with 3dDiff because you aren't outputting a dset, but instead numbers.

Please let us know how that seems.  (I expect we will make some more 3dEdu* progs, too, eventually.)

jbteves:
I suppose this may be a silly question, but, are THD_3dim* structs not constrained to be only 3 dimensions? I'd seen that used in other places but it wasn't clear to me if they could be 4D series, for example. I'll take a closer look at the edu program this week, that's a neat addition.

mrneont:
The THD_get_voxel() function (see thd_loaddblk.c) has the following syntax:
`float THD_get_voxel( THD_3dim_dataset *dset , int ijk , int ival )`
... where ijk is the index of element (e.g., voxel or surface node) and ival is the index of the volume (of a 3D or 4D set---in the former case, it would be 0).  This means this function can be applied to get values from 3D volumes or surfaces (or even *.1D files, if you transpose them after input).  So, your 3dDiff program could compare 3D volumes, surfaces, and more---even 4D files, if you looped over volumes (and you could output N values for two dsets of N subbricks).

jbteves:
Thanks for all the help! Heavily updated in the latest commit, hopefully this is closer to what you're looking for.

jbteves:
@pmolfese proposed in a different discussion that I could do both, either positions implying "left" and "right," OR if the flags exist then use them. He also proposed using `-dset1` and `-dset2`, though I wonder if `-a` and `-b` à la `3dcalc` might also work.

mrneont:
Hi, @jbteves -

To your question about naming inputs:  indeed, switching from -left and -right makes sense to me.  Whether to go for, say:
A) "all files at end of command line" 
B) explicit opts per file (-a/-b/... or -dset1/-dset2/...)
C) single argument that can take one or more files (e.g., "-infiles ..", like 3dmask_tool uses)
D) other
kind of depends on the future expanse of this program.

I might suggest B or C.  The appeal of "-a/-b/..." (or "-dset1/dset2/...) is brevity for 2 files, easy expandability for a few more, with possibilities for communicating text output about them by referring to each by their letter (or number).  Also, if the program expands, someday you might just add a "-infiles .." argument, to allow for lots of input datasets.  So, by the end of this paragraph, something from "B" for the moment, and possible expansion/change to "C" in the future.

Separate note:  the testing framework has finally been fixed (by the magical Yarik), and the shorter-term issue with CMake is also resolved in the current master branch.  So, I might suggest that you pull the master branch into this one---hopefully, there shouldn't really be any conflicts (maybe just minor Makefile stuff).

--pt

jbteves:
@mrneont adjusted for option (B), using `-a`/`-b` instead of the more verbose `-dset1`/`-dset2`. Last Q: right now, I mirror the behavior of `diff` where if everything matches, it quietly succeeds. Any thoughts on whether I should retain that or switch to stating that they match?

mrneont:
@jbteves : I might have thought the output to be (integer) 0 if there is no difference, and some kind of nonzero if there were any difference?  

Alternatively, 2 valued output:
+ if no difference, output: 0 0
+ if difference, output: 1 [the max or sum of diffs?]

Another consideration:  if someone puts in a 4D dataset of N subvolumes, are you aiming to output just 1 number, or N numbers?

jbteves:
So in the examples, right now you get verbal outputs rather than just numbers. If the grids diverge, for example, you get a message. When you have non-zero number of diverging _total_ voxels (regardless of 3D or 4D), it reports the percentage for simplicity. I could just modify the message to say "0 of N voxels diverge" which will closely match the statement I print for divergence.

mrneont:
@jbteves :

I see.  From a scripting/programming point of view, having either all or the [0]th output be a clear "yes" or "no" seems appealing to me.  This could take the form:
1 25 16777216
... to show that yes, there is a difference (the leading 1) in 25 out of 16777216 voxels, and
0 0 16777216
... to show that no, there is no difference (leading 0) in 0 out of 16777216 voxels.  Note how the 2nd zero is still present, even if unnecessary, in order to keep the same number of outputs regardless of whether there is a difference.  Also, there is no text here (and even all a single type---integer), so it's easier to read in.  I think these features would help implementing this program in a script.  From this information, someone could readily calculate percentages, if needed, whether with a shell script:
```
#!/bin/tcsh

set vals = `3dDiff -a DSET1 -b DSET2`
if ( ${vals[1] ) then
    set perc = `echo "scale=10; ${vals[2]}/${vals[3]}" | bc`
    echo "++ The percentage diff is:  ${perc}"
else
    echo "++ There were no differences among the ${vals[3]} voxels."
endif
```
... or using the wacky Python that the kids these days are so crazy about.

--pt

jbteves:
Well if you want the output from a programming POV, I would think it would be significantly easier to use the function I added to `thd_mismatch.c`. Right now `3dDfiff` basically prints out stuff like this without the need for a script.
I could add a mode with something like `-numbers_only`, which would just print out a series of numbers indicating various agreements/disagreements, if you think that would be useful, but again these are much more accessible in the C code already, so I viewed this as a reporting tool instead.
BTW, if I wanted to make Python bindings for the new function in this build,  is that a straightforward process or no?

mrneont:
@jbteves 

OK, I just ran the current code, and get no output for -a/-b dsets being the same, and this example for differencing:
```
$ 3dDiff -a mask.auto.nii.gz -b mask.auto_ALL1.nii.gz
++ Images diverge: 146997 of 211200 voxels disagree (69.601%)
```
So, sure, this is informative for a viewer.  I might change "voxels" to "elements, and "disagree" to "differ", and you might want to spit out the tolerance, too, even if unset:
++ Datasets differ: 146997 of 211200 elements (69.601%) differ at tolerance=[value]
Perhaps you might as well have friendly text for no difference:
`++ Datasets do NOT differ: 0 of 211200 elements differ at tolerance=[value]`

I would definitely add a "-programmy_output" option flag, for stark output noted above.  Or, have the brutalist output of only numbers be the default, and have a "-tell_me_nicely" (or "-descriptive") option to get more descriptive output.

re. Python bindings:  I am ignorant in how to do this, but I know it is possible, if programmed a certain (unknown-to-me-at-this-moment) way in C.

jbteves:
@mrneont I'm not really happy with how needlessly complex I made this, but here is a version that has "brutalist" output with documentation on what that means.

mrneont:
@jbteves -

OK, that looks good.  Are you sure you want to be in the business of evaluating header differences with this program?  I think I would take the approach of ensuring that the two input dsets are on the same grid (in the deep sense of `3dinfo -same_grid ...`) and returning "fail" and nothing else if those headers don't match.  If they *did* match, then you could proceed with everything else.  This would shorten the output checklist, too.

Basically, I would think users should use 3dinfo to compare the header information (it's already good at that), and then this 3dDiff program would be about voxelwise/elementwise differences.  If the datasets have the same matrix dimensions but different placements in space (due to *any* part of the [q,s]form matrix), I wouldn't think a voxelwise/elementwise comparison would be possible/meaningful.  That kind of partitioning of comparisons (header for 3dinfo, data/brick for 3dDiff) seems useful to me.

--pt

jbteves:
Hm, that's an interesting thought. Right now what happens is it indicates that the differences are impossible/meaningless by returning a -1 for the voxel counts. The outputs have grown quite unwieldly already, and in theory this program could grow still (via things like adding a `-mask` option, a `-scale` option, etc.) So would your ideal version of this, given the tools in `3dinfo`, look something like:
- Run 3dinfo to check header differences
- If there are header differences I detect, simply stop and tell the user "Hey, check out 3dinfo, you probably want to check with the following options..."
- If there are no discernible difference, count them up and then just return a triplet of information or single message? (summary differs/doesn't differ, total differing elements, total elements compared)

mrneont:
There is a function to check for dataset mismatches.  So, let's say you have 2 dsets, dset_A and dset_B.  The first thing I would do after loading them in would be to check :

```
// load in A and B
dset_A = THD_open_dataset(input_nameA);
if( (dset_A == NULL ))
   ERROR_exit("Can't open dataset '%s'", input_nameA);
DSET_load(dset_A); CHECK_LOAD_ERROR(dset_A);

dset_B = THD_open_dataset(input_nameB);
if( (dset_B == NULL ))
   ERROR_exit("Can't open dataset '%s'", input_nameB);
DSET_load(dset_B); CHECK_LOAD_ERROR(dset_B);

// check for mismatch
if( THD_dataset_mismatch( dset_A , dset_B ) ) {
   [output your message/output, whether in default/gentle or optional/brutal form]
   [delete/free dset_A and dset_B]
   [exit program in whatever fashion you are ending]
}

[... all the code to do the dataset testing/comparison that you want, because you survived the fireswamp of mismatched datasets]
```

In this way, your output should be simpler, rather than the current 8 numbers:
```
Brutalist output is useful for scripting. The output is an 8-integer 
array with the following elements:
	0:	Summary (-1 failure, 0 no difference, 1 some difference)
	1:	Dimensions (the number of differing dimensions)
	2:	Grid Spacing (0 if no difference, 1 if some)
	3:	Image Center (0 if no difference, 1 if some)
	4:	Obliquity (the angle of difference)
	5:	Orientation (0 if no difference, 1 if some)
	6:	Elements (total differing elements, -1 if cannot)
	7:	Total Elements (total elements compared, -1 if cannot)
```
... you can get rid of items 1, 2, 3, 4, 5.  So, you get the benefit of simpler reporting (only 3 numbers needed to output).  Basically, getting a "-1" as the first output can mean 'some kind of header difference---dear User, if you are unsure or surprised, prithee check "3dinfo -same_grid ..."' (or some other non-brutalist message).

If you add in a mask at some point, sure then that just would add a second comparison---check that dset_A and dset_B aren't mismatched, then check either of them with dset_mask.

--pt

jbteves:
There is a third option that I probably should have considered sooner, not sure how you feel about this:
- Add something like `-val_ndiff` to 3dinfo (there's already `-val_diff`), which basically does what 3dDiff does
Pros: adds to the existing tool that people are most likely to look at, reduces code bloat since the parsing infrastructure, etc., are already in 3dino
Cons: can't do things like masking and scaling.

pmolfese:
I personally like how there are usually multiple ways to do things in AFNI. So can I vote for both?

jbteves:
Well in theory both isn't too difficult, because part of this patch is the addition of THD_count_diffs, which just returns the total number of diffs if possible. Most of the logic is dealing with checks surrounding that small calculation.

jbteves:
Maybe dumb question: there is no macro for "total elements" that I can find in `3ddata.h`. I'm using `DSET_NVOX * DSET_NVALS`, but are there cases where that is incorrect?

jbteves:
Hm, I'd wondered about that myself since I feel like you could easily overflow that in my program. I'm also running into an issue trying to build my modified 3dinfo:
```
/bin/rm -f 3dinfo
/usr/bin/gcc -O2  -DDARWIN -m64 -Wcomment -Wformat -DUSE_TRACING -DHAVE_XDBE -DDONT_USE_MCW_MALLOC  -o 3dinfo 3dinfo.o -L. -L/usr/local/lib -L/opt/X11/lib -L/usr/lib -L/usr/lib -Wl,-x -Wl,-multiply_defined -Wl,warning -Wl,-bind_at_load  -lmri -lf2c -lXm -ljpeg.9 -lXt -lXft -lfontconfig -lpng16 -liconv -lXmu -lXp -lXpm -lXext -lX11 -lz -lexpat  -lm -lc
Undefined symbols for architecture x86_64:
  "_THD_count_diffs", referenced from:
      _main in 3dinfo.o
  "_space_to_NIFTI_code", referenced from:
      _main in 3dinfo.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [3dinfo] Error 1
```
While I realize this must mean I'm not correctly getting my new THD_count_diffs function included in libmri, I'm reasonably sure I haven't mangled `space_to_NIFTI_code`. Any thoughts on what the issue could be? I'm realizing now also that `3dDiff` shouldn't need to include the thd_mismatch object file since that should be in the library, so this is probably a local setup issue but I'm not sure.

jbteves:
I'm realizing that it's me not building libmri like I'm supposed to, sorry about that.

jbteves:
@afni-rickr would it be okay to use a long integer for my functions, or is that opening a can of worms the team would not like to deal with?

afni-rickr:
@jbteves long int is not reliable for being 64-bit, so we have gone with int64_t.  But that has the irritation of using it in a printf requires a PRId64 macro, or else there are warnings on one system or another (e.g. using %ld).  Just some fun to ponder...

afni-rickr:
For libmri, most programs have that as a dependency of course, be we have not bothered to list it for every one in Makefile.INCLUDE.  Adding that would also work.  But the dependency on thd_mismatch.o is a more localized one that is good.  I often build a program via  `make libmri.so PROGRAM` (as you did) since the dependency is not there.

mrneont:
Hi-

Slightly late to this party, which has been going strong this morning!

I think it would be better to just focus on 3dDiff, rather than mirror this functionality in 3dinfo, too.  At present, it seems possible to have have a simple option in 3dinfo, but I imagine the desired functionality of 3dDiff will quickly grow---as soon as you mention adding a mask (which seems a natural addition), then this no longer is in the realm of 3dinfo.  Then, you are stuck with awkwardly duplicating, or branching, or something else that can be avoided.  I am quite surprised that 3dinfo has -val_diff, actually... though I aaaaalways learn surprising things about the code base.

Many programs in AFNI work volume by volume---I suspect this would be a handy model here.  Rather than going for Nvox * Nval, when contemplating multivolume data, why not just treat each 3D volume separately?  The output could have N rows, then, one for each:
```
0   0  100000
0   0  100000
1 20  100000
```
You could also have a separate option for just one, summary row, either keeping the same format or adding a 4th column (number of volumes):
```
$ 3dDiff -a dset1.nii.gz -b dset2.nii.gz -diff_across_time
1   20  100000 3
```
... where now the 4th column is the number of volumes (in lieu of 3 columns, with the final value being 300000).  Even if there is only 1 volume, it could be there.  It isn't so wonderful changing the format output, perhaps (3 columns of numbers to 4), but at least if the user adds that option explicitly, then they can know what to expect.  

It also seems more meaningful to me to leave the Nvox and Nvalue info separately reported---can be useful to know, and most Nvox and Nvals are not so easily to decouple.
 
--pt

jbteves:
I'm actually quite confused by it as well, it struck me as an odd thing to include, yet when I was going hunting for header checking flags, it was right there alongside `-same_grid`. I like the idea of offering a tabular representation of the diffs, and an optional way to one-line it. This seems like a satisfactory way to avoid `printf` macros, which is deeply, deeply preferable from a newbie developer point of view... Additionally, you could make a more complete summary, such as:
```
Found differences in X of N volumes (P%), with total differing voxels ranging from A elements (a%) to B elements (b%)
```
So I'll work on making those changes.

Another potentially dumb question: why not add the LIBMRI_OBJ dependency to those files? It's already present in many of the makefile lines and proofs it a bit against future me. I don't trust that guy, he has a checkered past with following instructions, even from (and perhaps especially from) himself.

@pmolfese other than liking having multiple ways to do things, do you have a compelling argument for me to keep that change in `3dinfo`? It did seem quite out of place, and the behavior is a little bit weird because you actually get the same number back twice since it's designed to potentially compare many files.

mrneont:
I will haaapppily defer libmri questions to @afni-rickr .

`3dinfo` is generally best for getting info about a single dataset, and often for header-related items.  There are some data-related ones (e.g., min and max values), but that is rare.  There are a couple cases for comparing multiple files---namely, in that section you have pointed to, with the semi-commonly-used -same_grid, which performs the role of THD_dataset_mismatch() in shell scripts.  But for the most part, data-querying functionality tends to reside elsewhere happily, because it often gets more complicated such as by adding a mask.  Perhaps even the `-*val_diff` options exist in `3dinfo` because a timetraveling or alterate-reality-hopping JBT put it there, before realizing that programming a new C progarm would be **even more** fun than editing an existin gone.

--pt

jbteves:
I'm thinking that there's actually still one issue, which is that if you want a one-line option you could still quite potentially overflow an integer with the differences, if there are many differences in an image. So given that, would you prefer that the one-line option retain the number of volumes and voxels as separate items, or to use a double and then truncate the decimal and print out the total number of elements compared, and then print it out as one item? (Though I suppose even this could be made into an option).

jbteves:
To clarify: since the total number of differing elements will overflow an integer, that definitely needs to be a double (which should accommodate even fairly astronomically sized datasets, since my understanding is that we should get at least 15 digits), should I just go ahead and do the same thing with the total elements compared to simplify the output, or do you think having the total volumes is still preferable?

mrneont:
So, the issue is that intead of having a small number of total diffs:
```
$ 3dDiff -a dset1.nii.gz -b dset2.nii.gz -diff_across_time
1   20  100000 1000
```
... it could try to look like:
```
$ 3dDiff -a dset1.nii.gz -b dset2.nii.gz -diff_across_time
1   5000000000  100000 50000
```
... and unfortunately the number of diffs is greater than `2**31`, so this will create an overflow problem?

That is a very large number of diffs (a high-res, 50k volume dataset) that would lead to that.  But you could guard against it in a few ways:
+ putting a ceiling value on the number of diffs (`2**31`, say)
+ outputting the number of diffs as 5.0e+9 (that is, jumping to %e for printing output)
+ outputting a -1 as the second arg, to signify that the number of differences are "a lot", in a technical sense.

--pt

afni-rickr:
Yes, $(LIBMRI_OBJ) could be specified as a dependency of most programs.  Please feel free to add it to yours.  Often a program is written without updates to the library, and so the author might not worry or think about it.
Indeed, THD_count_diffs() should probably return int64_t, rather than int.

jbteves:
@mrneont @afni-rickr I think this version addresses everything we discussed via call yesterday except for adding a mask. What do you think?

mrneont:
There was a build failure in the checks here (and on my comp, too), because the code put into `3dinfo.c` is now broken by the updates to `int THD_count_diffs()`.  Namely, there is an extra argument in the function now (the dreaded `ival`):
```
int THD_count_diffs(THD_3dim_dataset *d1, THD_3dim_dataset *d2, int ival, float eps) {
```

I think removing the changes to 3dinfo.c will be necessary.  As we discussed, that functionality will only get left behind with other additions (e.g., the mask, which will surely appear at some point).

Will try building again on my comp with older `3dinfo.c`...

mrneont:
@jbteves : I still don't get how to make a PR to your codebase here, so I will email you a 3dDiff.c with a slightly updated help, as internal check for the various display opts.

Also, I think the quiet mode has an issue---I don't think the following output is correct (comparing identical volume, should be 0 in the output):
```
$ 3dDiff -a epi_r1+orig'[0]' -b epi_r1+orig'[0]' -q 
1
```

--pt

mrneont:
OK, to fix the quiet "-q" output, I think the if condition there should look like:
```
    if ( quiet ) {
        printf("%lld\n", ( total_elements_differing != 0 ));
    }
```
so that 1 is output if `total_elements_differing` does not equal zero, and 0 is output if `total_elements_differing` does equal zero.

jbteves:
I think this should fix the issues above, thanks for the suggestions!

jbteves:
@pmolfese sorry but we're going to pull this out of 3dinfo, it will just get to be a headache in the long run I think.

mrneont:
> OK, to fix the quiet "-q" output, I think the if condition there should look like:
> 
> ```
>     if ( quiet ) {
>         printf("%d\n", ( total_elements_differing != 0 ));
>     }
> ```
> 
> so that 1 is output if `total_elements_differing` does not equal zero, and 0 is output if `total_elements_differing` does equal zero.

@jbteves : I think this comment got lost in the thread;  I am pretty sure that the logic operator here should be `!=`, not `==`.

So I presumptuously put that in via this commit d20da95 (because I still don't see how to easily make a request to a PR).

jbteves:
Gah, I forgot to push the commit that had that change, it didn't get missed. Your commit is identical, however, except the comment.

jbteves:
I can show you on Zoom how to do a PR to a PR sometime, though in this case I'd recommend using "suggestion mode" instead.

mrneont:
Sounds good, thanks.  

I searched for "github comments", but that didn't give the "github suggestions" functionality that I had been looking for...  

Also happy to learn about the Zeno's PRs (PR of a PR of a PR of a PR, infinitely approaching but neeeever actually reaching the finalized code).

jbteves:
Just google "Josh's living nightmares," should pop right up.

To make a suggestion:
- Go to files changed
- Click on the plus sign next to the line you want to make a suggestion on (or use Shift and click on multiple lines, which turn yellow, and _then_ click the plus sign)
- Click "Make a suggestion," which will plop the current version in a block surrounded by backticks
- Replace it with your suggestion

src/3dDiff.c
```diff
@@ -0,0 +1,345 @@
...
+    set_obliquity_report(0); /* We'll check that ourselves below */
+    ds1 = THD_open_dataset(a_fname);
+    if ( ds1 == NULL ) ERROR_exit("Can't open dataset %s", a_fname);
+    DSET_load(ds1); CHECK_LOAD_ERROR(ds1);
+    ds2 = THD_open_dataset(b_fname);
+    if ( ds2 == NULL ) ERROR_exit("Can't open dataset %s", b_fname);
+    DSET_load(ds2); CHECK_LOAD_ERROR(ds2);
+    if ( THD_dataset_mismatch( ds1, ds2 ) ) {
+        ERROR_exit("Mismatch between dsets!\n");
+    }
```
mrneont, line 219: I wonder if the text output here should differ based on the user's selected mode.  For example, if the user ran `3dDiff -1 ...`, then it would be nice to have a `-1` output here, rather than this failure?

For example, within this if condition of a mismatch, each different format for providing this information can be used, similar to output in a non-mismatch case:

```
if ( quiet ) { 
    printf("-1\n");
    exit(1);
}
if ( tabular ) {
    printf("-1 -1 -1 -1\n");
    exit(1);
}
if ( report || tps_report ) {
    ERROR_exit("Mismatch between dsets!\n");
}
```

This is useful because in quiet or brutalist modes, the user only expects a single number or 4 numbers output, respectively---so that should be given to them.

src/3dDiff.c
```diff
@@ -0,0 +1,345 @@
...
+
+    /* Do the reporting */
+    if ( tabular ) {
+        for (int i = 0; i < nt; ++i ) {
+            printf(
+                "%*d:\t%d\n",
+                (int)ceil(log10(nt)), i, counts[i]
+            );
+        }
+    }
```
mrneont, line 251: I don't quite understand tabular output.  

```
$ 3dDiff -a epi_r1+orig'[0]' -b epi_r1+orig'[0]' -tabular
0:	0
0 0 211200 0.00000

$ 3dDiff -a epi_r1+orig'[0]' -b epi_r1+orig'[1]' -tabular
0:	144312
1 144312 211200 0.68330

$ 3dDiff -a epi_r1+orig'[0,1]' -b epi_r1+orig'[2,3]' -tabular
0:	144312
1:	160266
1 2 2 1.00000
```
In the single volume case, the initial zero (before the colon) is confusing until I see the multivolume case---it might be a bit hard for the user to get the main 0/1 output correct, because of the multiline issue.  Perhaps having the brutalist/summary line first would make the most sense, and then one has each [i+1]th line of output for each [i]th volume.  If that change isn't appealing, the current functionality should probably be spelled out more in the help, so user's don't utilize the wrong value.

jbteves: I think I'll spell this out in the help. I'm a little concerned about putting that line first, because in practice that will always be the last thing that the user looks at rather than the first for 4D data of any normal size.

src/3dDiff.c
jbteves: That makes sense, I'll patch that shortly.

jbteves:
While I was waiting on cmake to pass, I was starting to work on applying a mask. I see in 3dEdu01_scale.c (most notably, lines 394-399), you immediately NULL out the masked array. Is this a multithreading issue? In my case, I'll want to apply the same mask across many time points and two different dsets, so it would be a pain to keep NULL-ing and re-allocating it.

jbteves:
For some reason I seem to be getting an off-by-one index error with the patch I'm about to push. We can iterate through the (n-1)st voxel fine, but on the final one I get an illegal hardware instruction for both my masked and unmasked data, which is driving me mildly crazy because I don't see quite what I'm doing wrong or different from before. Pushing in case someone has a chance to look.

src/3dDiff.c
jbteves: In the most recent commit, the `diff3d_crash` macro handles this.

jbteves:
For further clarification: I get output like this
```
Iteration 126975 of 126976
zsh: illegal hardware instruction  /Users/tevesjb/repositories/afni/src/3dDiff -a $f -b $g
```
Where the statement in question is printing the current `ijk` being passed to THD_get_voxel OR being pulled from the double arrays, and the second value is `DSET_NVOX(ds)`.
The failure modes in the masked and unmasked cases are identical. I'm assuming there's something very simple that I'm overlooking but I can't seem to figure it out.

afni-rickr:
Hi @jbteves , `counts` needs to be allocated (maybe calloc) for the number of volumes.

jbteves:
Thanks @afni-rickr that was but one of many allocations that got just totally mangled when I rewrote the code to include a mask. I think this should work much, much better.

mrneont:
@jbteves : would it be a fair tradeoff if a couple of the tps_report_signoffs moved to closing messages, rather than stayed in said report?

jbteves:
I'm not sure I understand quite what you mean; at the moment it's all just printf statements. Is there some equivalent to `INFO_message` for closing messages? I could also strike the messages entirely, it was my own joke with myself about how somebody could ask for random and meaningless options.

mrneont:
@jbteves :

Closing message texts go into a separate file.  That makes sense about their origin, it just seemed like they probably shouldn't be in the final outputs here....

jbteves:
Should I also pull off the "TPS Report" heading and make the option something like "-long_report" instead?

mrneont:
"-long_report" does sound more intuitive.

mrneont:
I will move a couple over...  The file in question is afni_startup_tips.h.

jbteves:
Yes, I see that prodigious file.

While we're at it, could I add some Hitchhiker's quotes?

Is there anything you'd like to see changed before merging other than the addition of more help text?

src/3dDiff.c
```diff
@@ -0,0 +1,407 @@
...
+    /* If there's a mask, validate and apply it to the dsets */
+    if ( mask_fname ) {
+        total_elements = 0;
+        THD_3dim_dataset * ds_mask = THD_open_dataset(mask_fname);
+        DSET_load(ds_mask); CHECK_LOAD_ERROR(ds_mask);
+        if ( THD_dataset_mismatch( ds1, ds_mask ) ) {
+            diff3d_crash("Mismatch between input and mask dsets!\n");
+        }
+        if ( DSET_NVOX(ds_mask) != nv ) {
+            diff3d_crash("WTF bro?");
```
mrneont, line 251: How about slightly different message here?

src/3dDiff.c
```diff
@@ -0,0 +1,407 @@
...
+"\n"
+"  -q               :(opt) quiet mode, indicate 0 for no differences and\n"
+"                          1 for differences. (aka \"Rick Mode\")\n"
+"  -tabular         :(opt) display only a table of differences, plus\n"
+"                          a summary line (the same one as -brutalist)\n"
+"  -brutalist       :(opt) display one-liner. The first number indicates\n"
+"                          whether there is a difference, the second number \n"
+"                          indicates how many elements (3D) or volumes (4D)\n"
+"                          were different, and the last number indicates the\n"
+"                          total number of elements/volumes compared.\n"
```
mrneont, line 75: these should each describe the output when things don't match---particularly for brutalist and quiet

For tabular, I would provide an explicit example of what you call the "table of differences" part looks like, because it will be hard for the user to parse that couple numbers, esp. when only 1 volume is input, probably.

src/3dDiff.c
```diff
@@ -0,0 +1,407 @@
...
+"\n"
+"===========================================================================\n"
+"\n"
+"See Also ~2~\n"
+"\n"
+"    If this program reports that the images cannot be element-wise compared,\n"
+"you can examine the header information with 3dinfo. In particular, check out\n"
+"the section, \"Options requiring dataset pairing at input\", most notably\n"
+"options starting with \"same\", for example, -same_grid.\n"
+"\n",
```
mrneont, line 88: How about another section with a couple examples?

src/3dDiff.c
```diff
@@ -0,0 +1,407 @@
...
+"                          indicates how many elements (3D) or volumes (4D)\n"
+"                          were different, and the last number indicates the\n"
+"                          total number of elements/volumes compared.\n"
+"  -long_report      :(opt) print a large report with lots of information.\n"
+"\n"
+"If no display options are used, a short message with a summary will print.\n"
+"\n"
+"===========================================================================\n"
+"\n"
+"See Also ~2~\n"
```
mrneont, line 82: How about: 
```
"Notes ~1~\n"

"Re. 'failed' comparisons ~2~\n"
```

src/3dDiff.c
```diff
@@ -0,0 +1,407 @@
...
+int help_3dDiff()
+{
+char * author = "JB Teves";
+
+printf(
+"\n"
+"This is a program to examine element-wise differences between two images.\n"
+"\n"
+"written by: %s\n"
+"\n"
```
mrneont, line 52: ->

```
"\n"
"Usage ~1~\n"
"\n"
```

src/3dDiff.c
jbteves: That isn't actually even necessary because I check for a mismatch above, it was during angry debugging time and I forgot to remove it.

mrneont: OK. And note that often (or always?) a mask will just have 1 volume, even when applied to a multi-volume dataset.

jbteves: Always, in this case, because I just select the first sub-brick. I can note that in the help.

jbteves:
@mrneont I believe that the help should be fairly comprehensive, I took some liberties from your suggestions; hopefully, that's okay.

mrneont (approved):
Looks good.

PR #314, opened 2021-11-24
SUMA: Fixed "a" key so that the axes are toggled in clipping plane mo…

afni-laurenpd:
…de and toggles attenuation of background otherwise.  While functionality seems right, there are a lot of error messages that act like warnings


src/SUMA/SUMA_clippingPlanes.c
```diff
@@ -131,7 +134,9 @@ Boolean toggleClippingPlaneMode(SUMA_SurfaceViewer *sv, Widget w, int *locallySe
         if (SUMAg_CF->clippingPlaneVerbose && SUMAg_CF->clippingPlaneVerbosityLevel>1)
             fprintf(stderr, "### Darken inactive clip planes\n");
         darkenInactiveClipPlaneSquares(*locallySelectedPlane);
+        /*
         lightenActiveClipPlaneSquare(*locallySelectedPlane);
+        /**/
```
afniHQ, line 139:
```suggestion
        /*
        lightenActiveClipPlaneSquare(*locallySelectedPlane);
        */
```

mrneont: this has now been done in this commit: 80a2191

PR #316, opened 2021-11-30
Edt

mrneont:
3dEulerDist:  a program for calculating the Euler Distance Transform (EDT) in C code.

This code calculates the Euclidean Distance Transform (EDT) for 3D
volumes following this nice, efficient algorithm, by Felzenszwalb
and Huttenlocher (2012;  FH2012):

Felzenszwalb PF, Huttenlocher DP (2012). Distance Transforms of
Sampled Functions. Theory of Computing 8:415-428.
https://cs.brown.edu/people/pfelzens/papers/dt-final.pdf

Another useful/illustrative resource about this is by Philip Rideout:

   https://prideout.net/blog/distance_fields/

The current code here extends/tweaks the FH2012 algorithm to a more
general case of having several different ROIs present, for running
in 3D (trivial extension), and for having voxels of non-unity and
non-isotropic lengths.  It does this by utilizing the fact that at
its very heart, the FH2012 algorithm works line by line and can even
be thought of as working boundary-by-boundary.

Here, the zero-valued "background" is also just treated like an ROI,
with one difference.  At a FOV boundary, the zero-valued
ROI/backgroud is treated as open, so that the EDT value at each
"zero" voxel is always to one of the shapes within the FOV.  For
nonzero ROIs, one can treat the FOV boundary *either* as an ROI edge
(EDT value there will be 1 edge length) *or* as being open.

This code follows the earlier lib_EDT.py library version written by PA Taylor and earlier distanceField program by P Lauren.

### DIFF
--- a/src/3dEulerDist.c
+++ b/src/3dEulerDist.c
@@ -195,52 +195,109 @@ int main(int argc, char *argv[]) {
 int run_EDT_3D( int comline, PARAMS_euler_dist opts,
                 int argc, char *argv[] )
 {
-   int i, j, k;
+   int i, j, k, idx;
    int nn;
+   int nx, ny, nz, nxy, nvox, nvals;
 	THD_3dim_dataset *dset_roi = NULL;          // input
 	THD_3dim_dataset *dset_edt = NULL;          // output
 
-   int nx, ny, nz, nvals;
+   float ***arr_dist = NULL;           // array that will hold dist values
    
+   ENTRY("run_EDT_3D");
+
+   dset_roi = THD_open_dataset(opts.input_name);
+   if( (dset_roi == NULL ))
+      ERROR_exit("Can't open dataset '%s'", opts.input_name);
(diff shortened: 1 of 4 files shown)


neurolabusc:
Excellent. Two related tools come to mind:

 - Python bindings for C code, [anisotropy optimizations](https://github.com/seung-lab/euclidean-distance-transform-3d)
 - [Tool to compute EDT for NIfTI images](https://github.com/neurolabusc/DistanceFields) with documentation that describes several useful applications.

mrneont:
Hi, @neurolabusc :  This program works for anisotropic voxels, as well as for NIFTI inputs.  It is quite fast, at present

I think I could just put a switch or two in having it work for 2D datasets (just so it stops trying to get distance along 3rd dimension), and likely 1D, as well.

I have a Python version of this which is used in a different project.

--pt

mrneont:
...and that https://github.com/neurolabusc/DistanceFields is a nice page!

PR #324, opened 2021-12-13
notes on thd_niftiread.c

afniHQ:
for testing things out: to go into ni_attr branch


src/src_notes/notes_thd_niftiread.txt
```diff
@@ -0,0 +1,67 @@
...
+               ATRNAME_ANATOMY_PARENT, ATRNAME_WARP_PARENT, ATRNAME_DATANAME
+            - markers
+               ATRNAME_MARKSXYZ, ATRNAME_MARKSLAB, ATRNAME_MARKSHELP
+            - warp (any warp - affine or tlrc_12)
+               ATRNAME_WARP_TYPE, ATRNAME_WARP_DATA
+               - or auto_tlrc warp (ATRNAME_WARP_DATA_3DWD_AF), from 3dWarpDrive
+            - ATRNAME_BRICK_STATS      : brick stats
+            - tagset
+               ATRNAME_TAGSET_NUM, ATRNAME_TAGSET_FLOATS, ATRNAME_TAGSET_LABELS
+            - sform                    : IJK_TO_DICOM_REAL
```
afniHQ, line 37: I like jello

src/thd_initdblk.c
```diff
@@ -1124,18 +1141,15 @@ ENTRY("THD_datablock_apply_atr") ;
...
-              
-              if( fabs(old) + fabs(new) != 0 ){
-                  fracdiff = fabs( old - new ) / ( fabs(old) + fabs(new) );
-                  printf("fracdiff %13.6f\n",fracdiff);
-              }
-                            
+              float fracdiff;
+
+              fracdiff = frac_diff(atr_flt->fl[0],
+                                   dset->daxes->ijk_to_dicom_real.m[0][0]);
```
mrneont, line 1149: I don't think the float var here should have the same name as the function---it makes my head hurt.  (Perhaps it should make the compiler's head hurt, too.)

src/thd_initdblk.c
```diff
@@ -722,6 +722,23 @@ int THD_WarpData_From_3dWarpDrive(THD_3dim_dataset *dset, ATR_float *atr_flt)
...
+float frac_diff( float old, float new );
+
+float frac_diff(float old, float new) {
+    float fracdiff;
+
+    if (fabs(old) + fabs(new) != 0) {
+        fracdiff = fabs(old - new) / (fabs(old) + fabs(new));
+    } else {
+        fracdiff = 0;
+    }
```
mrneont, line 736: How about perhaps:

```
    float fracdiff, denom;

    denom = fabs(old) + fabs(new);

    if (denom) {
        fracdiff = 2 * fabs(old - new) / denom;
    } else {
        fracdiff = 0.;
    }
```

... which calculates the denominator only once? A factor of 2 in the numerator makes sense to me, from the point of view of considering the denominator to be the average of old and new.

src/thd_initdblk.c
```diff
@@ -1140,60 +1140,36 @@ ENTRY("THD_datablock_apply_atr") ;
...
       if(atr_flt) {
-          if( validate ){               /* disco change */
-
-              /* just testing here */
-              printf("ijk xx: %13.6s %13.6s %13.6s\n", "old", "new", "diff?");
-
-              int label_c = 0, count_o = 0, count_i = 0;
+        
+          if (validate) { /* disco change */
+              // should I free these vaiables at the end of the if statement?
```
afni-rickr, line 1145: A free is only needed when using a corresponding malloc/calloc/realloc.

PR #344, opened 2022-02-01
.docker/afni_dev_base.dockerfile: tweak curl because of Python 3.6 ve…

mrneont:
…r being used

### DIFF
--- a/.docker/afni_dev_base.dockerfile
+++ b/.docker/afni_dev_base.dockerfile
@@ -142,7 +142,10 @@ ENV PATH="/opt/cmake/cmake-${CMAKE_VER}-Linux-x86_64/bin:$PATH"
 
 RUN mkdir $PYTHONUSERBASE
 # Add some more test dependencies
-RUN curl -fsSL https://bootstrap.pypa.io/get-pip.py \
+# [PT: Feb 1, 2022] CircleCI wanted the following change, because we
+# specifically use Python 3.6 (alternatively, could update the ver of Python
+# below?)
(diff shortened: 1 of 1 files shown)


mrneont:
@afni-rickr : how does this seem to fix current CircleCI woes, such as here:
https://github.com/afni/afni/commit/b1e979563d6001d9ef04615bd4355198ca784b9c
and here:
https://github.com/afni/afni/pull/343/commits/4a9e5efd6b2ae6a5ad5c53ae3fbff3f148ce6622
?

Alternatively, we could relax/change the Python version to be 3.7 as opposed to the specified 3.6, but the above seems like a smaller change.

afni-rickr:
Alternatively, maybe we should test upping the version, maybe to 3.8 or 3.9.
Is this failing for all branches now (aside from this one with your fix)?

mrneont:
Yes, it appears to be failing on all branches built recently.

I am OK with updating the version of Python---I don't see why it would have to be frozen at v3.6

mrneont:
@afni-rickr : Python v3.9-dev does not exist, so trying for v3.8-dev.

mrneont:
Oddly, even specifying Python ver 3.8-dev failed with an error message about Python 3.6 not being compatible with the `curl -fsSL ...` comment. Sooo, trying just Python 3.8 (no dev) to see what happens.

mrneont:
@afni-rickr : I am defeated about how to get this Python version to really be above 3.6.  Even specifying other versions clearly still leads to Python v3.6 being used, and the original error message.

PR #348, opened 2022-02-04
Test.build.zss.040222

zsaad:


src/scripts_install/@MakeLabelTable
```diff
@@ -1,10 +1,12 @@
 #!/bin/tcsh
+onintr END
 
 @global_parse `basename $0` "$*" ; if ($status) exit 0
 
+set tmpdir = ./
```
mrneont, line 6: this is fine, but I think it could actually be just:
```
set tmpdir = .
```

src/scripts_install/@MakeLabelTable
```diff
@@ -1,10 +1,12 @@
...
 set stat = 0
 set RNS = `3dnewid -fun`
```
mrneont, line 8: I wonder if 
```
set RNS = "tmp_mlt_`3dnewid -fun`"
```
might be slightly preferable here?  That way, there is some consistent start to the filename, before the random chars.

PR #361, opened 2022-03-25
ENH - updated code to be able to read out the atlas label table

schoffelen:
Context: the Fieldtrip matlab toolbox allows for reading in atlases from various formats. Currently, there is hard coded support for the TT+tlrc atlas in (https://github.com/fieldtrip/fieldtrip/blob/master/fileio/ft_read_atlas.m) but I'd rather have this a bit more generic. It seems possible to decode from the files' *.HEAD the necessary information, without the need to rely on a hard coded list of labels/values. Of course I can easily solve this on our end, but perhaps it's more futureproof to solve it at the level of reading in the data into MATLAB. This PR attemps to resolve this.


mrneont:
Hi, @schoffelen : Thanks for contributing this to the repo.  We don't really do a lot with the Matlab codes here (as you might have noticed...), but it is good to know they are still useful.

Would you have an example or two of the commands for which you are adding functionality, just for our own information (and perhaps which could be added in a help)?

Thanks.

schoffelen:
Dear Paul @mrneont, I indeed noticed that the matlab-code does not seem to heavily used / actively developed :). Initially I PR'ed into Zaid's afni_matlab repo, but I thought it would be potentially more useful to PR closest to the source. 
My point indeed is that I can come up with a local workaround, i.e. interpret the *.HEAD header information on our end (i.e. adding some stuff to our local version of the file readers) but perhaps it's better on the long run to synchronize on both ends.
Sure I can provide some test code or demo the functionality. Also, no problem for me to update the functions' documentation a bit. Will do.

mrneont:
That all sounds great, thanks.

schoffelen:
I am not sure based on what the CI fails, but below is a chunk of matlab code that should at least show that the changes I suggest don't case a massive fail of the reading functionality. I tested the stuff on the afni_atlas_distr set of files. Please let me know whether you'd need some additional information from my end.


% test script for updated BrikInfo.m code

% the datadir contains the unzipped afni_atlases_distr.tgz
datadir = '/Users/jansch/Downloads/atlases_latest';
f = dir(fullfile(datadir, '*.BRIK.gz'));

err = false(numel(f),1);
for k = 1:numel(f)
  fname = fullfile(f(k).folder, f(k).name);
  [err(k,1), info{k}] = BrikInfo(fname);
end

% check that the reading did not yield errors
assert(all(err==0));

% the fieldtrip code in https://github.com/schoffelen/fieldtrip/tree/afni
% should be able to read in all images without error
ok = true;
try
  for k = 1:numel(f)
    fname = fullfile(f(k).folder, f(k).name);
    x{k} = ft_read_mri(fname);
  end
catch
  ok = false;
end
assert(ok==1);

hastable = false(numel(f),1);
for k = 1:numel(info)
  if isfield(info{k}, 'ATLAS_LABEL_TABLE')
    hastable(k) = true;
  end
end

% the fieldtrip code should be able to read in all atlases without error
ok = true;
try
  for k = find(hastable)'
    fname = fullfile(f(k).folder, f(k).name);
    y{k} = ft_read_atlas(fname);
  end
catch
  ok = false;
end
assert(ok==1);

PR #383, opened 2022-06-21
open_apqc.py: new prog; lib_apqc_html.py: updated javascript

hanayik:
Adding these files at the recommendation of @mrneont. These are the result of our successful hacking at the OHBM Brainhack.


afni-rickr:
Hi @hanayik.  I think the lib_apqc_html.py updates should also go under the afnipy directory.  Thanks!

hanayik:
@afni-rickr, no prob. I have moved it. This should now be reflected in the PR.

afni-rickr:
Gads, and I just noticed that open_apqc.py has a main(), so it belongs under scripts.  Sorry for not looking more closely to begin with...

hanayik:
Haha, ok. So the location of files is precisely the opposite of what Paul instructed :). BTW, this is likely a work in progress. We are planning to also add [NiiVue](https://github.com/niivue/niivue) to the generated QC reports.

mrneont:
Ugh, sorry, I did indeed reverse the directories to put those in...  I blame the cars driving on the wrong side of the road here, obviously.

Issue #390, opened 2022-07-29
[Question] Is Linux ARM64 a supported architecture?


markjens:
Hello,

I’d like to ask whether afni is supported on Linux/Mac ARM64 systems ?
I see `src/Makefiles*` for Mac M1 any many others, but none of them look appropriate for Linux ARM64.

Thank you!

afni-rickr:
We have no such device at the moment, but could look into it.  What laptop/desktop/other device(s) are you picturing running it on?

neurolabusc:
@markjens AFNI does compile for MacOS ARM64 systems. Simply use the [Makefile.macos_11_ARM_clang](https://github.com/afni/afni/blob/master/src/Makefile.macos_11_ARM_clang). The Apple [M1 has outstanding floating point performance](https://github.com/neurolabusc/AppleSiliconForNeuroimaging). 

I have not tried compiling AFNI for a Linux ARM64. I suspect this will require a new LINUX makefile that uses clang. Most non-Apple inexpensive ARM computers (like the [Pinebook](https://www.pine64.org/pinebook-pro/) use CPUs that do not have great floating point performance. However, I think that the Oracle and Amazon cloud ARM CPUs should be very capable. 

One option to evaluate this would be to use the in-built virtualization of the latest MacOS to run Linux. For example, [this video](https://developer.apple.com/videos/play/wwdc2022/10002/) details how you can run Linux as a guest on a M1 running MacOS, with an implementation [here](https://developer.apple.com/documentation/virtualization/running_gui_linux_in_a_virtual_machine_on_a_mac). This would allow you to test the ARM Linux build scripts before you deploy them onto the cloud.

neurolabusc:
@markjens yes it is possible, with the (solveable) caveats listed below.

@afni-rickr and @afni-dglen perhaps we can work together to clean up my rough solution. Oracle is graciously providing development time on their [80-core 3 GHz Ampere Altra A1](https://www.oracle.com/cloud/compute/arm/) systems (with bare metal allocation allowing a 2-socket 160 core and 1TB of RAM). You can request a server via [brainhack_cloud](https://github.com/brainhackorg/brainhack_cloud/issues). I am happy to walk you through the setup. You can select your desired Linux image including Debian-based Ubuntu or Fedora-based CentOS or Oracle. For my testing I used Oracle 8.6.

There are several caveats:

[...]

PR #396, opened 2022-08-16
Fix Makefile.linux_ubuntu_22_ARM_clang syntax

markjens:
It seems someone's editor broke it by formatting it (max line length).

Without these fixes the build was failing with errors like:
```
$ make itall
Makefile:43: *** missing separator.  Stop.
```


neurolabusc:
@markjens please update this with my latest modifications for [issue 390](https://github.com/afni/afni/issues/390) which automatically detects clang/gcc, support both x86-64 and ARM, and is tested with Ubuntu 18.04-22.04. I would suggest naming this file `Makefile.linux_ubuntu_22`

afni-rickr:
I did not notice this thread/pull request until the comment from @neurolabusc .  The updated Makefile is already in as Makefile.linux_ubuntu_22_64.  Sorry for the confusion.

PR #403, opened 2022-08-31
tighten text/json processing

afni-rickr:


mrneont (approved):
This cleans things up nicely, @afni-rickr ---thanks!

PR #404, opened 2022-09-02
Write err

afni-rickr:
Trap NIFTI write errors, and any in 3dcopy/3drefit.


jbteves:
FYI I see some spots in niftiread that should also trigger non-zero returns:
L826 ("if not a group, fail")

afni-rickr:
That is in the context of a NIML extension not containing a group element (maybe it is empty).  It is debatable what to do in that case, but the choice allowing it to continue reading the NIFTI dataset seems reasonable.  Then it becomes a consistency question with other lines, I guess.

PR #409, opened 2022-10-08
lib_ap_examples.py: format via lib_format_cmd_str.py

mrneont:


mrneont:
@afni-rickr : following your coding suggestion, the AP help examples can be formattable like this now.

Have mildly tweaked some of the left-padding and linewidth (but those can certainly be readjusted back to how they were).

Even tried to use correct indentation levels in the file!

mrneont:
(@afni-rickr : A very minor point: I also am not sure if the windent kwarg in the command_string() method in ~line 258 was being used?)

afni-rickr:
Thanks, windent was intended to be used for that 14, now 8.

PR #413, opened 2022-11-08
Add Makefile for MacOS 13 Ventura

pmolfese:
Still testing, but initial tests appear to work.

1. Some tidying up
2. Deleted most of the symbolic link requirements (thanks homebrew)
3. Rename output
4. Add libomp specific include and lib directories to linker for -lomp.


afni-rickr:
Great!  Maybe that can go under other_builds?  Thanks!

pmolfese:
Happy to have it live wherever...  Is there a way of distinguishing between what goes in the main directory vs. other_builds? 

Seems like this would be immediately used by anyone buying a new Mac who wants to run AFNI native on ARM. The Makefile.macos_11_ARM_clang works for MacOS 11 and 12. NIMH will be fully MacOS 12 in a month (Intel or ARM). Other institutes and extramural are rolling MacOS 13 since that's what comes pre-installed on new Macs.

afni-rickr:
The partitioning method, for whatever it is worth, has been to put Makefiles in the main directory if they are used with our actual build systems, and otherwise under other_builds.  Of course it does not feel like a nice split...

afni-rickr:
Thanks!

PR #419, opened 2022-11-16
Update @afni_R_package_install subscript for R versions > 4.0

pmolfese:
Fixes issues with installing Shiny packages in demo on R versions > 4. 

@discoraj does this look okay?

### DIFF
--- a/src/shiny/misc/OmicCircos_pkg_install.R
+++ b/src/shiny/misc/OmicCircos_pkg_install.R
@@ -7,7 +7,11 @@
 ## get R version 
 r.ver <- R.Version()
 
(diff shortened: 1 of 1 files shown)


pmolfese:
Tagging @afni-discoraj as well!

afni-rickr:
Sorry you have not gotten a reply on this @pmolfese.   Sure, it seems good (given my lack of R skills), though R > 4 is no longer handled.  That could be saved for when v 5 comes out...

afni-rickr:
Okay, I'll just merge.  If anyone wants to address the non-existent v5, that can be done later...  :)

PR #421, opened 2022-11-29
Fix SUMA_INCLUDE_PATH for the universal Makefile

martin-g:
1) Remove wrong path - `glib-2.0/include`
2) Add include path for aarch64 - `-I/usr/lib/aarch64-linux-gnu/glib-2.0/include`

This is one of the changes suggested by @markjens at https://github.com/afni/afni/issues/390#issuecomment-1219148589. It seems it was never committed to the repository.


afni-rickr:
Thanks, @martin-g!

PR #428, opened 2022-12-20
archivotron: Dramatic refactor to use only dictionaries

jbteves:
@mrneont sorry for the delay but here's a working solution for only using dicts and using the property decorator to just compute whatever information is needed. Additionally, higher layers in the hierarchy do not need to know implementation details for the lower layers, the constructor handles that for them. I also set verbosity to be a module-level global, and constructors no longer require many parameters to be passed in since so much can be re-computed cheaply. I know a bit of the terminology needs to be hammered out, and so does some functionality. Additionally, some parts of this can probably be turned into functions since they're repeated. Lastly, right now error reporting is not well considered.

### DIFF
--- a/src/python_scripts/afnipy/lib_ap_bids.py
+++ b/src/python_scripts/afnipy/lib_ap_bids.py
@@ -4,9 +4,14 @@
 
 # -------------------------------------------------------------------------
 #
-# A rough draft at storing data from BIDS datasets.  The goal of this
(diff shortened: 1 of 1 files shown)


jbteves:
@afni-rickr may also be of interest to you

mrneont:
this work was merged in in separate/co-authored commit #424 .

PR #432, opened 2023-01-17
Powell NEWUOA 

afni-dglen:
Powell NEWUOA would provide NaNs (not a number) if a constraint min=max  for any parameter. While not a typical use case, it can be a test case and was causing problems with a specific test.

align_epi_anat.py would fail on the resample step if the anat and EPI datasets (dset1,2) had different view space inputs - one was ORIG, the other TLRC (tlrc view) (MNI, waxrat, ... any non-ORIG space). The view is now set for the resample to match the master, the anatomical dataset. That view is also applied appropriately in the subsequent step for the  EPI skullstripping/automasking step too.

### DIFF
--- a/src/python_scripts/scripts/align_epi_anat.py
+++ b/src/python_scripts/scripts/align_epi_anat.py
@@ -2957,11 +2957,12 @@ class RegWrap:
    def resample_epi(  self, e=None, resample_opt="", prefix="temp_rs", \
         subbrick=""):
       o = afni_name(prefix)
+      o.view = ps.anat_ns.view
+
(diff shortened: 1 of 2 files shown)


mrneont (approved):
These seem fine to me.  

In simplex.c, you are preventing dividing by zero---fine.  Only other modification might be to have a wider window in the if-condition, e.g.:
```
EPS_PSIZ = 1.e-7;
 if( fabs(N_psiz[ii]) < EPS_PSIZ) dv[ii] = 0.0;
```
... but what you have seems good as a start and should likely be added.

PR #440, opened 2023-02-03
Add 3dSqueakyToy

jbteves:


afni-rickr:
Nice.  :)

jbteves:
I think I didn't quite configure the makefile correctly, do you know how I mangled it?

afni-rickr:
Hi @jbteves This looks okay to me and builds fine.  If you are concerned with CircleCI, I think we need to force that to use a more current macos system, so that is failing everywhere right now.

PR #447, opened 2023-02-13
[FIX] lop$gltList indexing variable

nmuncy:
Change indexing variable `ii` to `n` at line 1625 to resolve likely copy-paste error.
Fixes the `Error in lop$gltList[[ii]] : subscript out of bounds` error that results when running 3dMVM with AFNI Version `AFNI_23.0.03 'Commodus'`.


afni-gangc:
Thanks for reporting this! It was indeed a typo. The fix will be in the next official release.

afni-rickr:
Merged.  Thanks @nmuncy!

PR #456, opened 2023-03-09
codespell (#350) without the last "do it all" commit

yarikoptic:
should be safer initial batch, "manual labor" part of #350.  I will keep updating it as well.  It must have green CI.


afni-rickr:
@yarikoptic Thanks for this tractable version.  It seems good, so think it should just be merged now (before it grows).  Are you okay with it going in at this point?

afni-rickr:
Reviewing the other PR comments, I will guess that you are fine with merging and will do so.  Thanks again!

yarikoptic:
Thank you @afni-rickr . The only gotcha is that now workflow would keep failing for every PR. Well, could be taken as encouragement ;-) truly - try codespell on a few fingers close to your heart - it does quite a great job.

afni-rickr:
Indeed, and I am attacking some files.  :)  We still need to figure out how to fix CI anyway, this just adds another failure.  :)

PR #458, opened 2023-03-11
rspell

afni-rickr:
more codespell updates


yarikoptic:
I have reviewed the diff -- looked good to me. And with the superpowers vested (and not taken away from me yet), since CI builds/tests pass (besides a known failing one) and I wanted to do a little more of codespelling -- I will merge this. Thanks @afni-rickr for joining my codespelling quest!

tip of trade: with the most recent  version of codespell  (2.2.4) you should get coloring up of the wrong word in interactive session. So for ambiguous fixes etc I recommend trying `codespell -i 3 -c -w` invocation ;)

PR #459, opened 2023-03-12
another round of codespellling.

yarikoptic:
Most are "manual" in a sense that they are either explicit sed invocation, or going with `codespell -i3 -c -w` on some files. Concentrated on few components, e.g. SUMA but didn't finish SUMA -- more to do. Decided to make a stop here for now. Accompanied with https://github.com/afni/afni_ci_test_data/pull/7

There is also typo fix in data structure attribute `collapsable`, but I do not expect it to be used outside of the project so should IMHO be ok to fix.


afni-rickr:
@yarikoptic I have spent quite a bit of time going through some of this over the weekend, in hopes that no one else would be doing the same thing.  The files I have updates have been reasonably scrutinized (though some change output text, and I believe 2 actually changed variable names).  So we will likely be getting conflicts...  We can ponder.
Thanks!

src/SUMA/SUMA_3dHist.c
```diff
@@ -621,7 +621,7 @@ int main(int argc, char **argv)
          if (!strcmp(Opt->feats->str[ii], "upvol")) {
             float vv;
             if (Opt->f1 < 0) {
-               SUMA_S_Err( "Bad news seeding voxvol");
+               SUMA_S_Err( "Bad news setting voxvol");
```
yarikoptic, line 624: @afni-rickr this one and below -- I took my form `setting`since I have not spotted anything about `seeding` in the code, but I could have been wrong -- please let me know if needs to be fixed back.

src/SUMA/SUMA_Color.c
```diff
@@ -3182,7 +3182,7 @@ SUMA_Boolean SUMA_ScaleToMap_Interactive (   SUMA_OVERLAYS *Sover )
    if (Opt->Clusterize && Opt->RecomputeClust) {
       if (!(nv = (float *)SUMA_calloc(SDSET_VECFILLED(Sover->dset_link),
                                           sizeof(float)))) {
-         SUMA_S_Crit("Failed to alloc11659ate!");
+         SUMA_S_Crit("Failed to allocate!");
```
yarikoptic, line 3185: @afni-rickr this typo was not introduced in 478e7b07c241ad45cd9c6bdc1b2defa6facfe294 on purpose e.g. as an expression of emotions or smth like that, right ? ;-)

afni-rickr: :) Thanks!

afni-rickr:
This seems great, thanks!

afni-rickr (approved):
Thanks!

afni-rickr:
@yarikoptic Okay, this is merged.  How about you let this sit for a week or so, and we will do some work on our end, would that be okay?  Then we won't be stepping on each other's feet so much.
Thanks again!

Issue #452, opened 2023-03-02
ConvertSurface creates atypical PLY format


neurolabusc:
ConvertSurface creates legal but unusual [PLY format images](https://web.archive.org/web/20161204152348/http://www.dcs.ed.ac.uk/teaching/cs4/www/graphics/Web/ply.html). Specifically, ConvertSurface inserts a property named `intensity` that is [set to 1 for each and every vertex](https://github.com/afni/afni/blob/db1b39e7251d4939e087cbfed1209000600a18f3/src/SUMA/SUMA_Surface_IO.c#L4226). Since there is no variability in this property, it adds no information, results in larger files, slows reading, and can confound minimal PLY readers that assume canonical PLY files. 

Here is an example of a PLY format header created by ConvertSurface that includes the wasteful `property uchar intensity`:

```
ply
format ascii 1.0
comment author: Greg Turk
obj_info random information
element vertex 32492
property float x
property float y
property float z
element face 64980
property uchar intensity
property list uchar int vertex_indices
end_header
```

afni-dglen:
Hi Chris,
I've fixed this problem. The code used was actually from the original PLY definition examples. I think the idea was that faces of surfaces could be assigned arbitrary properties. That seems a useful feature, but it was never implemented, and the value of 1 was just a placeholder.
We rarely use the ply surfaces nowadays. How do you use them?
Thanks for letting us know. The fix will be in the distribution after I fix some of the other compiler warnings in these files while I'm looking at this.

neurolabusc:
@afni-dglen thanks. I do like binary PLY as it is faster, smaller and retains precision better than ASCII-based Wavefront OBJ. Since it re-uses vertices, it is far superior to STL. Therefore, from my perspective it is the best triangular mesh format that is well supported by many tools outside our domain Despite being a classic format, the simplicity of PLY keeps it on the Pareto frontier, as seen in this [JavaScript comparison](https://github.com/NeuroJSON/MeshFormatsJS).

PR #482, opened 2023-05-05
ply surfaces without intensity, clang march fix and compiler warnings

afni-dglen:
This set of changes is in response to the issue of atypical ply surfaces. Intensity is an unnecessary element in the faces property that gets repeated with values of 1 through the output file. Also a collection of small fixes to remove warnings on build in the relevant file. @afni-rickr can you take a look and verify. There was a minor include update for the marching cubes for clang compatibility.
https://github.com/afni/afni/issues/452

### DIFF
--- a/src/SUMA/SUMA_IsoSurfaceFunc.c
+++ b/src/SUMA/SUMA_IsoSurfaceFunc.c
@@ -10,6 +10,7 @@ http://www-sop.inria.fr/prisme/personnel/Thomas.Lewiner/JGT.pdf
 
 #include "SUMA_suma.h"
 #include "MarchingCubes/MarchingCubes.h"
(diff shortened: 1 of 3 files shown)


src/SUMA/SUMA_IsoSurfaceFunc.c
```diff
@@ -10,6 +10,7 @@ in Journal of Graphics Tools.
 
 #include "SUMA_suma.h"
 #include "MarchingCubes/MarchingCubes.h"
+#include "MarchingCubes/mc_ziad.h"
```
afni-rickr, line 13: Did this include come from the master branch or did you make a separate update?  If separate, this will be a merge conflict.  No biggie though.

PR #486, opened 2023-05-26
ASC: add check (with 'please fix'-ability) for Xvfb

mrneont:
@afni-rickr : how does this seem to include?  Xvfb is needed for `@chauffeur_afni`, and hence APQC HTML, etc.


afni-rickr:
That might be better off just added to check_list, a few lines up.  If that were done though, it does not seem to have a method for getting the version.

PR #494, opened 2023-06-09
note who is building

afni-rickr:


afni-rickr:
Okay, maybe the CI test failure was just a hiccup.

PR #527, opened 2023-09-27
Changed CalcRanksForReHo to properly calculate ties at the end

kevindtran1:
The CalcRanksForReHo function had a small edge case bug where the assigned ranks would not be adjusted for ties if these ties occurred at the end of the sorted time course. The value of LENTIE may be greater than 0 but sorted[m]!=sorted[m-1] will be false at the end of the loop. This code change will fix the issue.


mrneont:
This does seem appropriate.  Thanks for pointing this fix out, @kevindtran1 !

PR #539, opened 2023-10-27
@chauffeur_afni func_resample logic error

kmorwood:
The input parameters of @chauffeur_afni for -func_resam expect one of NN, Li, Cu or Bk. The logic in the script reports an error because it is ORing those tests. This means you cannot use this functionality at all because it is impossible for the value to be all of those at once.

A simple inversion of the logical OR to AND and all is well.

We have tested this in our project, where we found the problem in the first place, and it resolves nicely.

### DIFF
--- a/src/scripts_install/@chauffeur_afni
+++ b/src/scripts_install/@chauffeur_afni
@@ -689,9 +689,9 @@ while ( $ac <= $#argv )
         if ( $ac >= $#argv ) goto FAIL_MISSING_ARG
         @ ac += 1
         set func_resam = "$argv[$ac]"
-        if ( ( "$func_resam" != "NN" ) || \
-             ( "$func_resam" != "Li" ) || \
-             ( "$func_resam" != "Cu" ) || \
+        if ( ( "$func_resam" != "NN" ) && \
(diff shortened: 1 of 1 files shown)


kmorwood:
There are no CI tests for the change we have made. I'm not sure why the circleci is failing.

mrneont:
Howdy, @kmorwood : Ah, totally, thanks for finding+squashing that bug.

The CircleCI woe is separate and longer running, related to a server crash we had here that affected our testing infrastructure.  We still need to get on fixing that.

Thanks again,
pt

kmorwood:
Strictly for education...what is your build cycle? (Aka: when would the next build of the system come out?)

afni-rickr:
Hi @kmorwood.  We actually ran a build last night (though our build macs are having emotional issues).
There isn't exactly a build cycle, except for "when we want to put out an update".  The average interval between official builds is about one week, but it varies quite a bit.

kmorwood:
Thank you for the update!

I probably should have added that I am using the docker container extensively. I am capable of building my own container. I just need to figure out which of the Dockerfile is used to build afni/afni_make_build. At a glance I'm not sure which to choose.

PR #541, opened 2023-11-04
p2dsetstat: now a C language version (with more helpful output text)

mrneont:


mrneont:
This was unnecessary... putting in other branch that was already started for this.

PR #560, opened 2024-01-11
Packaging additions

TheChymera:
Old work by @leej3 and some additions by me. Work in progress for the time being

### DIFF
--- a/cmake/FindGTS.cmake
+++ b/cmake/FindGTS.cmake
@@ -1,5 +1,5 @@
 #[=======================================================================[.rst:
-Findgts
+FindGTS
 ---------
 
 Find the native gts headers and library.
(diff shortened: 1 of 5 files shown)


afni-rickr:
I would not go the route of renaming the library, as it really is the same library.  It seems that cmake system has a -DUSE_SYSTEM_F2C=ON flag that you could use if it is already installed.
Turning off whirlgif is really easy (comment it out in src/CMakeLists.txt).  However there is a test to make sure that it is being made, which fails at that point.  It would have to be removed from the test, too.  We can ponder that.

TheChymera:
@afni-rickr this should be ready for review now. The test failure I think is on the test runner side ( `worker 'gw0' crashed while running 'scripts/test_3dTstat.py::test_3dTstat_basic[cvarinv]'` ).

TheChymera:
@afni-rickr would you be interested to merge? this allows me to use CMake for the Gentoo build, and it works.

afni-rickr:
That's great @TheChymera, thanks!

PR #565, opened 2024-02-02
Suma20240202

afni-laurenpd:
New branch.  Alpha seems to work and no toy example crashes observed on Ubuntu system


mrneont (commented):
@OtagoHarbour : I think the new line 38 could use updating:
`	D$SUMA_DIR/riveSuma -com  viewer_cont -key '.'`
->
`	$SUMA_DIR/DriveSuma -com  viewer_cont -key '.'`

PR #568, opened 2024-02-07
aprs

afni-rickr:
Add regress to -radial_correlate_blocks.
But need to update DEFAULT example names. 
And then modify afni_proc.py to handle example names with underscores, rather than spaces.


afni-rickr:
Hmmm, this may have conflicts with my other branch.  Such is life...

PR #620, opened 2024-04-25
add -uvar

afni-rickr:


afni-rickr:
still need to pass through gssrs

PR #633, opened 2024-05-17
Homebrew gcc-14 new requirements on C code

afni-dglen:
gcc-14 is the latest default version for homebrew with `brew install gcc`. That latest version of gcc imposes some C code updates like requiring int declarations for functions that return integers. Updates to ignore some warnings with -Wno-implicit-int (f2c code built from Fortran would need to be updated with each conversion) and a couple small other declaration fixes solves the complaints of this version of the gcc compiler.

The libjpeg.a update only occurs in the configure command script it uses to determine compiler properties and not in the AFNI source code.

The libXmHTML code required two small changes for consistent type declarations.

### DIFF
--- a/src/XmHTML/lib/Motif/motif.c
+++ b/src/XmHTML/lib/Motif/motif.c
@@ -281,7 +281,7 @@ DestroyImage(XImage *image)
 }
 
 static unsigned long
-GetPixelWrapper(XImage *ximage, Dimension x, Dimension y)
+GetPixelWrapper(XImage *ximage, _XtDimension  x, _XtDimension  y)
 {
 	return(XGetPixel(ximage, x, y));
 }
--- a/src/3dExchange.c
+++ b/src/3dExchange.c
@@ -157,9 +157,9 @@ int main( int argc , char * argv[] )
       fac = DSET_BRICK_FACTOR(dset, i); /* get scale factor for each sub-brik*/
       if(fac==0.0) fac=1.0;
       nvox = data_im->nvox;             /* number of voxels in the sub-brik */
-
-      outfar = outsar = outbar = mri_data_pointer(out_data_im);
-
+      outbar = (byte *) mri_data_pointer(out_data_im);
+      outfar = mri_data_pointer(out_data_im);
+      outsar = mri_data_pointer(out_data_im);
       badsb_type = 0;
       switch(data_im->kind){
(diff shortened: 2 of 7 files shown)


src/XmHTML/lib/Motif/motif.c
```diff
@@ -281,7 +281,7 @@ DestroyImage(XImage *image)
 }
 
 static unsigned long
-GetPixelWrapper(XImage *ximage, Dimension x, Dimension y)
+GetPixelWrapper(XImage *ximage, _XtDimension  x, _XtDimension  y)
 {
```
afni-rickr, line 285: It might not matter, but I would expect _XtPosition to be more a more appropriate type for x and y.

afni-dglen: That might be. I used `_XtDimension` because that is the default prototype used in tka.h for the GetPixel function, but this is confusing because the original Dimension is an `unsigned int`  defined in miniparse.h. In Xutil.h (part of the X11 library, provided on the Mac via XQuartz), XGetPixel uses either `int` directly or the x and y are defined within a macro that calls get_pixel (Xlib.h), which uses `int` for x and y.

Intrinsic.h in the X11 library specifies _XtDimension as either `Dimension` (defined as `unsigned int`) or `int` depending on whether it is being used with a "wide" variable definition. They also say not to use _XtDimension and similar variable types in client code - not sure if that means XmHTML code. 

So either `_XtDimension`, `int` or `unsigned int` seem to be reasonable choices.

afni-rickr: Cool, given the prototype, _XtDimension does seem good.

afni-rickr:
Can you make sure your two SUMA changes are reflected in the 0327 branch?  The feel free to merge, thanks.

afni-dglen (commented):
Included all changes in the suma20240327 branch too. Tested  build against FT05 NL afni_proc script from class data, and all looked good except for numpy2 issues in the QC HTML generation on Mac. Those are outside this pull request. Rerunning with conda-managed numpy1 produced expected QC HTML pages with appropriate visual results.
@afni-rickr - can you do final approval?

mrneont:
What be the QC HTML issues?  Let me know if I should take a look.

afni-rickr (approved):

PR #635, opened 2024-05-24
3dTsplit4D: add -bids_deriv opt, for potentially-growing name management

mrneont:


mrneont:
@afni-rickr : here is a simple tweak to the existing program. It does the current job for adjusting the output names appropriately, from map_ap_derivs.py's standpoint.  But we can certainly head a different route.

afni-rickr:
Cool, though did you want to check for '-' as the previous character?  Or maybe ( ! isalnum() )?

afni-rickr:
@mrneont How does this seem?  So currently, -bids would not be needed, but it is fine to leave in for future updates.

PR #645, opened 2024-05-29
Sum quick fix20240528

afni-laurenpd:
Fixed bugs where program crashed when "A" box widget not present or when datatype does not support colored clipping planes


afni-laurenpd:
Fix clipping plane bugs and prevent menu stretching

PR #672, opened 2024-07-16
flat and patch states

afni-dglen:


afni-dglen (commented):
That looks good to me.

PR #673, opened 2024-07-16
circleci: update docker version in config, bc previous ver now deprec…

mrneont:
…ated


mrneont:
This commit+update is because this build update (https://github.com/afni/afni/commit/8b432aa575a3305b205312a2a4d45f42994233c1) resulted in a failure in on circleci testing, stating the erstwhile docker version 20.10.11 was now deprecated. So, this updated is to "docker23". See the comment in the commit about where this syntax came from.

PR #682, opened 2024-08-07
ROI2dataset added help to match nodes of another dataset

pmolfese:
Borrowed text from help file on ConvertDset to remind users of ROI2dataset can pad an ROI to match another Dset directly. Thanks @afni-dglen for the tip.


pmolfese:
Actually looks like this behavior may not transfer over from one tool to another.

mrneont:
Just quickly looking at the code for SUMA/SUMA_ROI2dataset.c, I don't see any option processing for `-pad_to_node`. What was the full command you ran?

pmolfese:
It's in ConvertDset but apparently not in ROI2dataset. I'll fiddle with copying functionality over once I get back from vacation! Switched PR back to draft.

mrneont:
Surely a vacation to not be working on programming---buon viaggio!

pmolfese:
Wow that was a lot of layers deep. So yes, _even though_ the option isn't in the SUMA_ROI2dataset.c file, it looks like all of the pad_to_node functionality exists in `AFNI_environ.c`. That behavior then gets carried over to any/all SUMA programs. So the change in the help is correct and warranted. 

There is a nuance in that I think that some of the R-based programs don't correctly set max_node correctly (verified with `3dinfo -max_node`) so even though they show a correctly R-to-L extent as the number of nodes, the header somehow doesn't carry over to max_node. 

Maybe @afni-rickr can let me know how to 3drefit or modify a surface .niml.dset or Gifti to set that parameter for copying.

pmolfese:
Full working command: ROI2dataset -prefix test.pm -input rh.roi.stg.niml.roi -pad_to_node d:ROI2.rh.asdset2.niml.dset -overwrite

mrneont:
Thanks for the code example, and for delving into the mysterious depths of SUMA.  I'm happy to merge this in. 

Do you have a dataset with the wrong output of `3dinfo -max_node NIML_DSET` to check on what is happening with that?

mrneont:
Rockin'.

PR #683, opened 2024-08-07
ffmpeg parameter deprecation

afni-dglen:
The ffmpeg -intra option was deprecated about 6 years ago and is now obsolete on homebrew's default version. Replaced the "-intra" with "-g 1".

### DIFF
--- a/src/imseq.c
+++ b/src/imseq.c
@@ -4931,7 +4931,7 @@ ENTRY("ISQ_saver_CB") ;
                pattrn=getenv("AFNI_MPEG_PATTERN")  ;
                if( pattrn == NULL ){                  /* 07 Apr 2009 */
                  /* by default use only I-frames */
-                 if( adup <= 1 ) pattrn="-intra";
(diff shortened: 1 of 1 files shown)


mrneont:
@afni-dglen : check Part 1 is done, successfully: I can confirm that on previous code, I could *not* save an MPG from suma, but after building your new code here I *could* do so.

Next to check on Linux-Ubuntu...

mrneont:
@afni-dglen : .... and check Part 2 is done, successfully: though I note that on Ubuntu-Linux, both the *old* code and the *new* code allowed me to successfully save an MPG from suma.  So, "no harm done" here, I believe.

Looks good to me, and thanks for doing this!

afni-dglen:
Thanks for checking. It's not surprising the -intra option works on older versions of ffmpeg, distributed by package managers for the different OS's. AFAICT, ffmpeg made this option obsolete in version 5. The version I am running on my M1 ARM Mac is version 7.0.1.  I think Fedora recently made ffmpeg available too (version 6.1.1  https://koji.fedoraproject.org/koji/packageinfo?packageID=35247)

PR #690, opened 2024-08-26
FindMyStats added just to keep track of changes

mrneont:


mrneont:
how about renaming to include "afni", someline like:
* `FindAfniStatsDset`
* `find_afni_stats_dset`
etc?

--pt

Issue #654, opened 2024-06-05
Failed to compile code due to conflicting between Boolean and RwcBoolean for Xinstric functions


runlevel5:
```
```
$ uname -ar
Linux shrimp-paste 6.8.11-300.fc40.ppc64le #1 SMP Mon May 27 14:48:15 UTC 2024 ppc64le GNU/Linux

$ gcc -v
Using built-in specs.
COLLECT_GCC=/usr/bin/gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/ppc64le-redhat-linux/14/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: ppc64le-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,objc,obj-c++,ada,go,m2,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-targets=powerpcle-linux --disable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --enable-libstdcxx-backtrace --with-libstdcxx-zoneinfo=/usr/share/zoneinfo --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --with-isl=/builddir/build/BUILD/gcc-14.1.1-20240522/obj-ppc64le-redhat-linux/isl-install --enable-offload-targets=nvptx-none --enable-offload-defaulted --without-cuda-driver --enable-gnu-indirect-function --enable-secureplt --with-long-double-128 --with-long-double-format=ieee --with-cpu-32=power8 --with-tune-32=power8 --with-cpu-64=power8 --with-tune-64=power8 --build=ppc64le-redhat-linux --with-build-config=bootstrap-lto --enable-link-serialization=1
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 14.1.1 20240522 (Red Hat 14.1.1-4) (GCC)
tle@shrimp-paste:~/Work/afni/build$ g++ -v
Using built-in specs.
COLLECT_GCC=/usr/bin/g++
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/ppc64le-redhat-linux/14/lto-wrapper
[...]

PR #704, opened 2024-09-24
RwcBoolean to char

afni-rickr:
Xt event handlers use char instead of unsigned char, and it should not really matter for a boolean type.
This was brought up in https://github.com/afni/afni/issues/654 .


afni-rickr:
Yeah, I forgot to pull the master in initially...

PR #706, opened 2024-09-30
escape the escapes

afni-rickr:
python is viewing the \ chars as escapes, so they should be passed through to the shell


mrneont:
@afni-rickr : My only complaint here is that this seems like a missed opportunity to call the commit message "The Great Escape (of Escapes)" or similar.

PR #714, opened 2024-10-11
SUMA: Attempt to fix bug when B key toggles background on and off

mrneont:


mrneont:
@afni-dglen and @afni-rickr say OK!

PR #719, opened 2024-10-24
ap.volreg no volreg

afni-rickr:


afni-rickr:
Also included a surf blur fix, to avoid conflicts that I would have generated.

PR #733, opened 2025-01-14
tcsh - maybe WIP

afni-rickr:
invoke script tcsh via env rather than path, so remove all -f and add checks when removing -e


afni-rickr:
The commits before "general tcsh via env" were to add some $status handling when -e was being removed.
After that, they were basically sed changes via a modified `@strip.whitespace`.
`@Align_Centers` has not yet been done.

afni-dglen:
This looks pretty good. @Align_Centers relies on the -ef error handling, even noted in the comments. I'm looking at the details.

afni-dglen:
I made the updates for @Align_Centers. There is an exising bug that I did not fix here - if using the combination of NIFTI input and an AFNI format prefix.
Other continued /bin/tcsh in shell scripts that generate other convenience shell scripts as output, usually for showing output  - @RetinoProc, @chauffeur_afni, @measure_bb_thick and a few others create tcsh scripts. They output instructions to run these with "tcsh ..." followed by the name of the script, so effectively they will use the one first in the path.

afni-dglen (commented):
As mentioned in comments, these changes are limited to the shebang in most of the cases, so we expect them to have no effect on the functionality. However, there are a large number of files affected, with some in fairly obscure shell scripts that haven't been touched in many years. We will wait for final approval from @mrneont .

PR #736, opened 2025-01-26
cbar_pbar.py: start of cbar_pbar cmd line prog

mrneont:


mrneont:
this code was directly merged in, as the new colorbar_tool.py and lib_cbar_tool.py, so will delete this branch now

PR #741, opened 2025-02-03
Roistats.format

afni-rickr:


afni-rickr:
@pmolfese How does this seem?  Thanks.

pmolfese:
Looks great to me! Tested with very small numbers and seems to be providing the correct info!

afni-rickr:
Great, thanks!

PR #772, opened 2025-03-25
align_epi_anat.py: 'mkdir' -> 'mkdir -p' (with backslashes)

mrneont:


mrneont:
Changing "mkdir" -> "\\mkdir -p" should both provide more consistent behavior across systems (from the escape slash) and avoid this unnecessary error message in many cases (from the `-p`):
```
mkdir: cannot create directory '<OUTDIR>': File exists
```

mrneont:
I also wonder if this program should have "-VERB" on in 3dAllineate by default? That puts a looot of text in the output, most of which seems pretty unnecessary to me?  A user could always add that with `-Allineate_opts -VERB` if desired?

mrneont:
@afni-dglen : it would be great to have your opinion on this minor-change PR, and the suggested other tweak to AEA here.

mrneont:
@afni-dglen has now confirmed that he is the happiest person in the world with this branch, so merging it in.

PR #773, opened 2025-03-25
@animal_warper: no -iwarp for auto_warp.py

mrneont:


mrneont:
@afni-dglen : it would be great to *also* have your opinion on this PR, which has non-trivial but hopefully mostly negligible change, around `-iwarp` usage.  On my test dset, the follower D99 atlas mapped to subject space was identical when using either the original AW or this modified version.

afni-dglen:
I expect this particular implementation of changes for warp inversion will give different results in many cases. The  "-iwarp" option here is only for 3dQwarp.  There, 3dQwarp usually handles nonlinear warps in the same space with source and target grids exactly the same with affine transformations  usually run beforehand. 

But 3dQwarp can do affine transformations too via its -allineate option. In that case, the regular WARP output includes the affine transformation with in the warp. The affine transformation, even when it's included in the output as a separate 1D file, is still integrated into the WARP file.  In the allineate case, we don't have the advantage of a nonlinear only part of the WARP by itself. That means inverting there doesn't really matter which we do. The iwarp of 3dQwarp probably inverts that transformation better, but I can't be sure. In this case, the INV(warp) idea may be riskier than the original.

mrneont:
Maybe we can chat about this in person. In this case of `@animal_warper`, the affine part is explicitly done separately by `align_epi_anat.py`, so there will be no `-Allineate` option handed to 3dQwarp via `auto_warp.py`, surely?  This seems the simpler case of just not having 3dQwarp create its own, lone inversion.

But I might (likely) be confused.  Assuming that the "here" is `@animal_warper`, which we are discussing updating here, I am not sure what "there" is.

afni-dglen:
I both mistyped my comment and misread the change. I had thought this was inside auto_warp.py instead of animal_warper. For animal_warper, this change is completely reasonable. Thanks for doing that.

mrneont:
Rockin'. And merging this PR, then. Thanks.

PR #792, opened 2025-05-05
Be bragging no colonhscopies

mrneont:


src/scripts_install/@update.afni.binaries
```diff
@@ -815,11 +815,11 @@ if ( $local_package != "" ) then
    endif
 
    # get the directory of the file
-   if ( $local_package == $local_package:h ) then
```
mrneont, line 818: when would this ever be true?  Just if `$local_package` is `.`, yes?  Is that the point of this if condition, @afni-rickr ?  Nothing wrong with it

src/scripts_install/@DiceMetric
```diff
@@ -171,7 +171,7 @@ if ("$prefix" !~ "") then
    echo "#Col.0 = 'ROI label in $base ($lll)'" >> $prefix.1D
    set cnt = 1
    foreach dset ($dsets)
-      set bp = $base:h
+      set bp = `dirname $base`
```
mrneont, line 174: This seems fine, just surprised that this is done inside the loop over `dsets`, since `$base` is independent of that?

zsaad:
True, does not need to be in the loop

afni-rickr:
Sorry, I did not see this conversation until my screen refreshed with a push.

src/scripts_install/@update.afni.binaries
afni-rickr: This is true if there is no '/' in the value.

PR #822, opened 2025-08-19
AP suggestion: handle/fail differently for polort<0 in 3dToutcount

mrneont:
@afni-rickr : 

What do you think about this change in handling polort values for 3dToutcount?  This was prompted by a case in which a user had `-regress_polort -1`, which created a failure later on down the road in the proc script execution.

Separately, 3dToutcount fails with polort <0, but with the wrong error message, so I am going to edit that in the master branch.

Basically, 3dToutcount cannot have a polort <0.  This puts a check for that earlier on, and fails if a user advertently or inadvertently creates a situation for that.  But there may be better ways to implement this in the code.  I have greatly expanded the number of lines in this function...

### DIFF
--- a/src/python_scripts/afnipy/db_mod.py
+++ b/src/python_scripts/afnipy/db_mod.py
@@ -906,8 +906,17 @@ def make_outlier_commands(proc, block):
     # set polort level
     val, err = proc.user_opts.get_type_opt(int, '-outlier_polort')
     if err: return
-    elif val != None and val >= 0: polort = val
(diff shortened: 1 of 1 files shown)


mrneont:
I also wonder if in line 908 the return should be:
```
return 1, ''
```
... instead of being empty?

afni-rickr:
Sure, yes to both of those (though the regress_polort error might do well to mention -outlier_polort).
Thanks!

PR #827, opened 2025-08-29
@afni_refacer_run: new 'cubeface' mode and opts

mrneont:


mrneont:
This is no longer needed, as it was superseded/added in creating afni_refacer2 (https://github.com/afni/afni/commit/75329b62d38acf3ad1505e341e918cf62348897d).

PR #829, opened 2025-09-04
afni_open: new '-s' opt, to open spreadsheets; prob needs more things…

mrneont:
… to search for


mrneont:
@afni-rickr : have added that functionality, thanks for the suggestion.

PR #835, opened 2025-09-15
Sort datasets in dataset choosers 

afni-dglen:
This update sorts datasets in the AFNI GUI dataset choosers like Underlay and Overlay choosers. Previously, datasets were ordered in an apocryphal order related to their dataset type and their data format with anatomical datasets first, followed by all other types. AFNI format was  included first and then NIFTI and other types of datasets.

The new orders allow for simple alphabetic, reverse alphabetic, dataset type (anatomical, echo planar, functional bucket - a little like the original, but includes NIFTI), or the default order the datasets are read in (close to the original method). I haven't preserved the real original sort, but that could be added in if someone wants that.

Tested with datasets with mixed spaces and formats. Read in new sessions, added new datasets after reading session, used AFNI_GLOBAL_SESSION.

We could make this an environment variable that can be changed within the GUI easily or a sort control within the chooser.

### DIFF
--- a/doc/README/README.environment
+++ b/doc/README/README.environment
@@ -368,6 +368,16 @@ the directory name are used for the display. If AFNI_SESSTRAIL
 is not set, then it is equivalent to setting it to 0 (which
 was the old method).
 
+---------------------------
+Variable: AFNI_SESSION_SORT
+---------------------------
+This is a string to set the sort order within the dataset choosers
+in the afni GUI. The allowed values are
+ALPHA    = alphabetical, case insensitive (A=a)
+REVALPHA = reverse alphabetical
+MRITYPE  = dataset type - anatomical, functional
+NONE     = previous default sorting of MRI type, then NIFTI, other types
+
 --------------------
 Variable: AFNI_HINTS
 --------------------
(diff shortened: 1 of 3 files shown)


mrneont:
Looks great to me. Each of the environment options worked as appropriate.  

I have AFNI_GLOBAL_SESSION on pointing to my directory of reference datasets, and each was sorted appropriately within its folder.

I also tried it with `afni -R -all_dsets`, and everything worked, sorting appropriately by directory name and then files.

It looks like it sorts:  numbers, then like "_", then alphabetical characters.  For alphabetical, I am running on Ubuntu, and it does not preferentially put all caps and then all lowercase, which is great;  it goes:  `AAA, aaa, BBB, bbb, CCC, ccc, ...`.

I haven't tried building it on a macOS yet.  Is the ordering the same as that?

Grazie *mille* for sorting this out!

afni-dglen:
Latest update uses toupper() for uppercase characters as intermediate. That gives sorting of numbers, alphabetic, underscore. The main motivation here is to put the underscore leading characters last because we use those only as intermediate file names, and these would typically appear only for exiting errors during processing.

mrneont:
@afni-dglen  : how about changing the description of "REVALPHA" to be "reverse of ALPHA", because using the term "alphabetical" doesn't really describe things fully?  We have more than just alphabet letters here; it is alphanumeric plus symbols.

PR #836, opened 2025-09-15
Crosshair gap size

afni-dglen:
Crosshair gap size can be set by default with an AFNI environment variable, AFNI_CROSSHAIR_GAP. Set this in the shell environment, on the command line via "-DAFNI_CROSSHAIR_GAP=nn" or in the .afnirc configuration file. Allowable values are between -1 and 19.

In response to this request on the AFNI messageboard:
https://discuss.afni.nimh.nih.gov/t/crosshair-color-and-gap-size-env-variable/3702

The current behavior is kept where new controllers take the default initial value rather than inheriting the gap size from the current controller. Should new controllers take their value from the previous controller setting, encouraging the same gap across all controllers, or take their initial value from the default (which can now come from the environment)? If a user updates the gap from 5 to 12, should a new controller be 12 or 5? Should gap size be automatically propagated across all controllers, so that if a user changes the gap size in one controller, the gap size changes in all controllers?

### DIFF
--- a/doc/README/README.environment
+++ b/doc/README/README.environment
@@ -1679,6 +1679,12 @@ range is 0 .. 0.05 (which will be very thick lines, I assure you).
 This variable was introduced in March 2015 for Corianne (if there is
 such a person).
 
+----------------------------------
+Variable: AFNI_CROSSHAIR_GAP
+----------------------------------
+This numeric variable lets you set the gap between the image viewer crosshairs.
+The default value is 5 voxels. Allowable range is from -1 (closed) to 19 voxels.
+
 -----------------------------
 Variable: AFNI_CROSSHAIRS_OFF
 -----------------------------
--- a/src/afni_history_dglen.c
+++ b/src/afni_history_dglen.c
@@ -53,6 +53,10 @@
 
 
 afni_history_struct dglen_history[] = {
+{ 15, SEP, 2025, DRG, "afni",
+     MICRO, TYPE_NEW_ENV,
+    "AFNI_XHAIR_GAP can be set to specify crosshair gap"
+},
 { 19, AUG, 2025, DRG, "3dXYZcat",
      MICRO, TYPE_BUG_FIX,
     "3dXYZcat generated extra empty volume in X direction"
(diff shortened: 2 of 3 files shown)


mrneont:
To these questions:

> Should new controllers take their value from the previous controller setting, encouraging the same gap across all controllers, or take their initial value from the default (which can now come from the environment)? If a user updates the gap from 5 to 12, should a new controller be 12 or 5? Should gap size be automatically propagated across all controllers, so that if a user changes the gap size in one controller, the gap size changes in all controllers?

I would make the default the value in each new controller, not the changed setting;  I think that is how all properties work, like crosshair color, posterior-is-left, etc. The user may well be opening the new GUI controller to load a different kind of data with different considerations, in which case the value need not propagate.  Users who are opening the new controller to look at things the same way can use the environment variable to make that behavior constant.

afni-dglen:
@afni-rickr I'm not sure I understand. I don't see that environment variable usage in master. I only see that in this xhairgap branch as AFNI_CROSSHAIR_GAP. In this branch, I did update README.environment. Where should I look? I could add a setting or a commented setting in AFNI.afnirc to match the current default of 5.

afni-rickr:
I think it is only crosshairgap in afni.c, and not in any other file..  There is no AFNI_ either, those variables are initialized via a macro and function call that prepends AFNI_.  But also, try running `afni -DAFNI_crosshairgap=0` from the master branch.

afni-rickr:
And hunting through this, I see the same applies for the xhair color.
`afni -DAFNI_crosshairgap=0 -DAFNI_ovcrosshair=11`

afni-dglen:
@afni-rickr Wow. That is arcane stuff. I followed that NAME2INT macro through the other files that it uses to see how it works. It can get the environment variable either from the shell environment, .afnirc, command line or even from .Xdefaults (after xrdb -load .Xdefaults with a line AFNI:crosshairgap 0). For the "regular" environment choices, it takes either the lowercase version of the "AFNI_crosshairgap" or all uppercase "AFNI_CROSSHAIRGAP" in its logic.

Okay, so I can close this and just put a note in the README.environment and AFNI.afnirc files so no one has to go through this again.

PR #915, opened 2026-07-24
3dShuffle

pmolfese:
Permutation testing program `3dShuffle`. 

Currently focuses on the first intended use case: single-brick datasets with matched subjects across repeated-measure conditions. Pairing is assumed by input order. The program supports labeled conditions, optional subject labels, directional contrasts, exact or random sign-flip permutations, voxelwise empirical p-values, and max-stat FWE correction. Output is a single bucket dataset with mean difference, observed paired t, uncorrected p/z, and FWE-corrected p/z bricks for each contrast.

```
3dShuffle                                \
  -conditions 2                          \
  -cond_labels endsleep longrec          \
  -input end_s01+tlrc end_s02+tlrc ...   \
  -input ltr_s01+tlrc ltr_s02+tlrc ...   \
  -contrast longrec endsleep             \
  -method signflip                       \
  -stat paired_ttest                     \
  -tails one                             \
  -mode exact                            \
  -prefix ltr_gt_end_perm
  ```


pmolfese:
Adds `3dShuffle` for doing permutation tests on a small number of subjects. `3dShuffle`, unlike other AFNI non-parametric tools, does not rank the data. It keeps the original voxel values, computes a t-style statistic, and asks whether that statistic is extreme under sign flips or group-label shuffles. The classic tools rank observations within each voxel and test whether rank patterns differ by group/condition.

mrneont:
Hi, @pmolfese : 

This looks like it is moving along nicely!  I was wondering about the tail-related calcs and calls.

I haven't done a deep read of the code, but I wondered if it would be useful to highlight some already built-in functionality to translate p-to-stat and stat-to-p in AFNI programs that might be useful.

This line in p2dsetstat.c shows how a stat value gets mapped to a p-value, with sidedness accounted for:
https://github.com/afni/afni/blob/8956c071ce3b919b4c051b16b5fce67a923893f9/src/p2dsetstat.c#L286
NB: just above it is is a check that the dset has a statcode attached to it, via `DSET_BRICK_STATCODE()`.  So, you could attach an appropriate statcode to the dset in question and make use of something like this?
This might be able to replace `p_to_signed_z()`?

In case it is useful, here is an example of the opposite case of p-value to stat:
https://github.com/afni/afni/blob/8956c071ce3b919b4c051b16b5fce67a923893f9/src/dsetstat2p.c#L286

Also, for selecting sidedness, it would be nice to have the keywods be "1sided", "2sided", "bisided", as those tend to be keywords used in various AFNI programs (the above ones, 3dClustSim, etc.) and so a bit more generally scriptable.

Again, these might not be relevant to the actual use case of the code, but I point them out just in case.

--pt

pmolfese:
Very helpful suggestions! I obviously found the DSET_ function but `THD_volume_pval_to_thresh` is super helpful in exactly this case (as you likely knew).

PR #924, opened 2026-08-12
1dCorrelate: update ziggurat alg file (and include testing in src_notes)

mrneont:


mrneont:
this branch is unneeded, because the contents of it were merged in from a separate zgauss2 branch that had a cleaner commit history (i.e., less back and forth of commits with changes of direction and fixes).

PR #928, opened 2026-08-13
Return the exit code from asynchronous command execution

### LINKED ISSUE #927: tests: non-zero exit codes are never detected (async executor returns None)
### Summary

`run_cmd` in `tests/afni_test_utils/tools.py` ends with a check that is meant to fail a test when the command it ran exited non-zero:

https://github.com/afni/afni/blob/0488c9ee5f277a66bd0b9a8ac1832ebc8fea6b2f/tests/afni_test_utils/tools.py#L555-L557

```python
    # Raise error if there was a non-zero exit code.
    if returncode:
        raise ValueError(f"{cmd}\n Command returned a non-zero exit code.")
```

That check can never fire on the default code path. `returncode` comes from `__execute_cmd_args_asynchronous`, which awaits only the two stream-reader tasks and then falls off the end of the function:

https://github.com/afni/afni/blob/0488c9ee5f277a66bd0b9a8ac1832ebc8fea6b2f/tests/afni_test_utils/tools.py#L337-L371

It never calls `p.wait()` and has no `return` statement, so it returns `None`, and `if returncode:` is always false. Since `use_asynchronous_execution` defaults to `True` (L423), this is the path essentially every test takes. The synchronous variant `__execute_cmd_args` returns `proc.returncode` correctly (L409), so only the asynchronous path is affected.

### Impact

No test can fail because of a non-zero exit status. A program can exit with an error, or die on a signal, and the test still passes as long as the output comparison passes or is skipped.

I ran into this concretely: while working on #925, `3dttest++` terminated with **SIGSEGV** (after being handed a truncated input file) and the test still reported `PASSED`. The only reason I noticed was that the run took 0.31s.

### Reproduction

Add a test module that runs a command which simply exits non-zero:

```python
from afni_test_utils import tools

data_paths = {}

def test_rc_probe_nonzero_exit(data):
    differ = tools.OutputDiffer(data, "false", merge_error_with_output=True)
    differ.run(skip_output_diff=True)
```

On current master this passes. It should fail with `ValueError: ... Command returned a non-zero exit code.`

### Proposed fix

#928 — await the process once both output streams have reached EOF and return its exit status. Because `watch()` has already drained both pipes at that point, the usual deadlock hazard around `wait()` with `PIPE` does not apply.

This is deliberately separate from #925/#926, and the two are less connected than my wording above first suggested. In #925 the command was never executed at all, so it never reached this check — an unused-local warning would have been the thing to catch that one (`rc` was assigned and never used). This issue is a different failure mode: commands that *do* run, and fail, without the test noticing.

Where they did meet is in verifying the #925 fix: with the test finally executing `3dttest++`, the program died on SIGSEGV and the test still reported `PASSED`. So this bug did not hide #925, but it does hide the next one.

cindykrafft:
Fixes #927.

`run_cmd` ends with a check that raises `ValueError` when a command exits non-zero, but that check could never fire on the default code path. `__execute_cmd_args_asynchronous` awaited only the stdout/stderr reader tasks; it never called `p.wait()` and had no `return` statement, so it returned `None`. `use_asynchronous_execution` defaults to `True`, so this is the path essentially every `OutputDiffer`-based test takes.

The effect is that no test could fail because of a non-zero exit status — a program could abort or die on a signal and the test still passed, as long as any output comparison passed or was skipped. The synchronous path already returns `proc.returncode` correctly; only the asynchronous path was affected.

The fix awaits the process after both output streams reach EOF and returns its exit status. `watch()` has already drained both pipes by that point, so the usual deadlock hazard around calling `wait()` with `PIPE` does not apply here. The timeout path is untouched — it still raises `TimeoutError` before reaching the new code.

### Verification

**The fix catches what it should.** With a probe module that runs a command exiting non-zero:

```python
def test_rc_probe_nonzero_exit(data):
    differ = tools.OutputDiffer(data, "false", merge_error_with_output=True)
    differ.run(skip_output_diff=True)
```

- before: `1 passed`
- after: `1 failed` — `ValueError: false\n Command returned a non-zero exit code.`

**No regressions.** Full suite run in `afni/afni_cmake_build`, same machine, same test data, with and without the change:

| | result |
|---|---|
| without fix | 1 failed, 141 passed, 28 skipped |
| with fix | 1 failed, 141 passed, 28 skipped |

The failure sets are identical. The one failure, `test_testing_script_functionality.py::test_run_cmd_timeout`, fails **before** the change as well, with the same exception at the same place (`TimeoutError` from `tools.py:371`, on the fourth `run_cmd` call). That test asserts on wall-clock deltas with a 0.2s time unit, and I was running an amd64 image emulated on an arm64 host, so I'd expect it to be environment-specific rather than a real defect — either way it is unrelated to this change.

I did not otherwise alter which tests pass, which was the main risk here: the check has been dormant for long enough that enabling it could have surfaced latent failures. On this suite it surfaced none.


afni-rickr:
Thanks!

PR #931, opened 2026-08-20
ENH: New Program - SurfClustSim

pmolfese:
# Add SurfClustSim: a native surface cluster simulator

Potentially replaces the tcsh pipeline generated by 
`slow_surf_clustsim.py` with a single C program, 
and adds `-acf` to `SurfFWHM`. 

## Why

`slow_surf_clustsim.py` is a script *generator* — At defaults (`niter 1000`,
`itersize 10`, 6 p-values): 100 `3dcalc`, 100 `3dVol2Surf`, 100 `SurfSmooth`,
**6000 `SurfClust`**, **6000 `awk`**. At `niter 10000` the last two become
60,000 process spawns each. Every `SurfClust` call re-reads the spec file and
rebuilds the entire surface graph to threshold one sub-brick; tabulation then
spawns one `awk` per file to read back a single number. The workload is
spawn-and-reload bound, not compute bound, and none of the four binaries has
OpenMP.

Useful for fMRI Surface analysis. Extra useful for my EEG/MEG source models
that are occasionally estimated at both subject and group levels.  Analyses now 
take minutes instead of hours. 

## What this does

Loads the surface once, works in memory, threads over simulations. Results
are highly comparable to `slow_surf_clustsim.py` under `-compat` and 
`-legacy_sided` options, ~ 0.2% difference in my testing, attributed to 
Monte Carlo simulation seeds as expected. 

**Smoothing modes:** `-compat` (default) reproduces the old script including its
adaptive FWHM search; `-fixed -sigma S -Niter K` pins it; `-acf a b c` matches a
requested autocorrelation instead of a FWHM.

**Sidedness:** `-1sided`, `-2sided`, `-bisided` as in `3dClustSim`, plus
`-legacy_sided` reproducing the old script's per-tail-alpha convention.

**`-multithresh`** — inspired by ETAC. Picking one cluster-forming threshold
before seeing the signal is arbitrary and changes the answer; this calibrates
the family jointly, so a result is significant if a cluster survives at ANY
threshold. One shared tail probability sets every cutoff (the "equitable" part),
solved by bisection. Validated out of sample — calibrated on one seed, applied
to 5000 fresh simulations from another:

| nominal | 0.10 | 0.05 | 0.02 | 0.01 |
|---|---|---|---|---|
| achieved | 0.1090 | 0.0518 | 0.0188 | 0.0096 |

Deliberately **not** called `-ETAC`: real ETAC draws its null from permuted
residuals of real data (where most of its advantage lies), supports richer
merits, calibrates over multiple blurs, and produces a spatial map. This has
none of those, and the help says so.

**`SurfFWHM -acf`** — estimates the ACF vs geodesic distance and fits AFNI's
mixed model `a*exp(-r²/2b²) + (1-a)*exp(-r/c)`. The surface counterpart of
`3dFWHMx -acf`, which has no surface equivalent today. A FWHM pins the *width*
of the autocorrelation, not its *shape*, and cluster extent at high thresholds
is governed by the tail — the mismatch that inflated cluster false positives in
the volume. Validated on fields with known answers: heat-smoothed white noise
fits `a=0.958` (nearly Gaussian), a two-scale mixture fits `a=0.533` (where real
fMRI sits; `3dFWHMx -help` quotes 0.578).

**`SurfClustSim -acf`** — uses `SurfFWHM -acf` output. A single heat kernel 
can't produce an exponential tail, so noise is a weighted sum of independent fields 
smoothed by differing numbers of passes; independent unit-variance components 
make the mixture's ACF the weighted sum of theirs, so the weights are a 
least-squares fit solved once at startup. The program then generates fields, 
measures them, and reports what it achieved. End to end:

| | a | b | c |
|---|---|---|---|
| measured from data | 0.5334 | 8.397 | 36.33 |
| regenerated | 0.5481 | 8.078 | 35.02 |

## Compatibility

`slow_surf_clustsim.py` is untouched and remains the reference. `-maxarea_1D`
writes raw max areas in the old `z.max.area.*` layout for comparison through
`quick.alpha.vals.py` (one difference: the old script omitted lines for empty
simulations, this writes an explicit 0 — pass `-niter` and alphas agree).
`SUMA_SurfFWHM.c` is purely additive, no existing line altered, no shared header
touched. `mri_fwhm.c` is untouched — the ACF fit reuses AFNI's own
`ACF_cluster_to_modelE()` via a 1-D pseudo-cluster.

## Tests

`tests/scripts/test_SurfClustSim.py` needs no test data — `CreateIcosahedron`
builds the surface, so no datalad dependency and no stored reference. All
assertions exact, not tolerance-based: sweep vs flood fill, determinism,
thread-count invariance, `-itersize` invariance in `-fixed`, `-itersize`
*dependence* in `-compat` (guards an inherited quirk deliberately), one line per
simulation, multi-threshold cutoffs never below per-threshold ones, and a slow
test that generated ACF matches the request.

## Known limitations

- **No noise floor for old-vs-new.** 851.20 vs 849.03 is 0.26%, almost certainly
  Monte Carlo noise, but that hasn't been established. Running the old script
  twice with different seeds would settle it.
- **`-compat` results depend on `-itersize`** — inherited, not new. `SurfSmooth`
  averaged FWHM across whatever columns it got and applied one `Niter` to all,
  so the old script's "block size (speed-up)" knob silently changed the null
  too. Reproduced on purpose; a test guards it.
- **Volume mode doesn't scale** — `vol2surf` isn't reentrant, so that path is
  serialized and its mapping rebuilt per block. Surface mode is unaffected.
- **`-acf` accuracy:** `a` within ~0.05; `b` and `c` looser (weakly identified
  when their component carries little weight).
- **Three kernels forked from SUMA** (HEAT_07 double + float, 1-difference FWHM),
 mostly to avoid possibly changing any behavior in other programs  by adding openmp calls.

### DIFF
--- a/src/CMakeLists_mri.txt
+++ b/src/CMakeLists_mri.txt
@@ -508,6 +508,11 @@ add_library(
   mri_write_angif.c
   mri_zeropad.c
   rcmat.c
+  # Gaussian deviate generators.  These live in a subdirectory and were added
+  # to MRI_OBJS in Makefile.INCLUDE but not here, so the CMake build had no
+  # definition for zgaussian2()/zgaussian2_init() and libmri failed to link.
+  zgaussian/zgaussian.c
+  zgaussian/ziggurat.c
 )
 target_link_libraries(mri_objs PRIVATE libheaders)
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,10 +94,17 @@ commands:
             # stop if running locally and it happens to exist...
             pre_existing=$(docker ps -aq --filter name=testing_afni)
             if [[ ! "" == "$pre_existing" ]];then docker stop testing_afni;docker rm testing_afni;fi
-            # Run the tests
+            # Run the tests. Temporarily disable errexit so a test failure
+            # doesn't abort the script before the docker cp below runs --
+            # otherwise the container (and its captured test output) is
+            # torn down with nothing ever copied back out.
+            set +e
             /tmp/src/afni/tests/run_afni_tests.py $TEST_ARGS
+            test_exit_code=$?
+            set -e
             # after application container finishes, copy artifacts directly from it
             docker cp -a testing_afni:/opt/afni/src/tests/afni_ci_test_data/. /tmp/src/afni/tests/afni_ci_test_data
+            exit $test_exit_code
 
   check_integrity_of_containers:
     description: "Check some of the permissions/user/PATH basics of the containers"
@@ -409,6 +416,13 @@ jobs:
           image_cache_version: << parameters.image_cache_version >>
       - run_afni_tests:
           build_type: << parameters.build_type >>
+      # Preserve each test's captured stdout/stderr as a downloadable
+      # artifact, even when the test step above fails, so a failure's
+      # actual output survives past the container being torn down.
+      - store_artifacts:
+          path: /tmp/src/afni/tests/afni_ci_test_data/output_of_tests
+          destination: test-output
+          when: always
       # Save test data as cache for subsequent runs if ideal is not already saved
       - save_cache:
           key: v<< parameters.test_data_cache_version >>-/tmp/test_data_version.txt-{{ checksum "/tmp/test_data_version.txt" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,10 @@ commands:
             set -e
             # after application container finishes, copy artifacts directly from it
             docker cp -a testing_afni:/opt/afni/src/tests/afni_ci_test_data/. /tmp/src/afni/tests/afni_ci_test_data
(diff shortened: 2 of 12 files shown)


src/CMakeLists_mri.txt
mrneont, line 1: Thanks for adding these to the cmake list! I wasn't sure how best to do that.

mrneont:
@pmolfese : Ha! very accurate commit name.

... and I'm sure I'll break cmake again!  Bwahaha.  Ignorance really _is_ bliss!

pmolfese:
> @pmolfese : Ha! very accurate commit name.
> 
> ... and I'm sure I'll break cmake again! Bwahaha. Ignorance really _is_ bliss!

I don't doubt it! Both filter and gaussian!  Pushing those together in a joint cherry pick, which was confusing enough that I finally just asked Claude to do it.

PR #938, opened 2026-08-24
Add preliminary support for Siemens XA Enhanced MR DICOMs

jberrebi:
Hi,
There seems to be need for Siemens Xa enhanced Dicom in real-time.

[https://discuss.afni.nimh.nih.gov/t/afni-realtime-siemens-sms-epi-error/7408/15](https://discuss.afni.nimh.nih.gov/t/afni-realtime-siemens-sms-epi-error/7408/15)

So, In December 2024/January 2025, I started to modify the source code for reading enhanced DICOMs. I found a routine able to read 7D Dicoms so I tried to use it in mri_read_dicom.c:

im = mri_new_7D_generic(nx,ny,nz,1,1,1,1, datum, g_dicom_ctrl.read_data);

then I modified or rather duplicated parts of the code in mri_read_dicom.c to a new case called "enhanced" that I kept as close as possible to "mosaic". I have added new entries in mri_dicom_elist.h for enhanced dicoms.

Then later I compared motion estimation from realtime afni (realtime dicoms Siemens XA, real-time afni with modified mri_read_dicoms and mri_dicom_elist.h) ) to offline afni and dicom that reconstructed the usual way. It seemed to be fine as reported in:

[https://discuss.afni.nimh.nih.gov/t/motion-parameters-afni-rt-with-dimon-vs-afni-after-dcm2niix/8934/9](https://discuss.afni.nimh.nih.gov/t/motion-parameters-afni-rt-with-dimon-vs-afni-after-dcm2niix/8934/9)

So now, I have adapted that patch to the latest version of afni. I have tested with phantom  and human being inside MRI and it works: motion estimation curve, fmri signal, display of the brain in realtime. The only issue is the following error in the terminal:
** DICOM ERROR: oversize length=1297920 in element (7fe0,0010)
** failed to read file '001_000023_000007_1.3.12.2.1107.5.2.43.166088.2026082418173622595921675.dcm' as dicom
6 ** DICOM ERROR: oversize length=1297920 in element (7fe0,0010)
** failed to read file '001_000023_000007_1.3.12.2.1107.5.2.43.166088.2026082418173622595921675.dcm' as dicom
But it seems that the dicom is not fully written. And then it must retry because it works "graphically". 

I don't know if it is the right way to go but as someone from abroad wrote to me that he was interested I wrote this patch.
Best Regards,
Jonathan

### DIFF
--- a/src/mri_dicom_elist.h
+++ b/src/mri_dicom_elist.h
@@ -72,7 +72,7 @@ static char *elist[] = {
  "5200 9230" ,  /* Per-frame Functional Groups Sequence */
  "0018 9074" ,  /* Frame Acquisition DateTime */
  "0018 9082" ,  /* Effective Echo Time */
- "0008 002A" ,  /* Acquisition DateTime */
+ "0008 002a" ,  /* Acquisition DateTime */
 
 
 NULL } ;
--- a/src/mri_dicom_elist.h
+++ b/src/mri_dicom_elist.h
@@ -42,28 +42,37 @@ static char *elist[] = {
  "0002 0010" ,  /* Transfer Syntax [RWC - 05 Jul 2006] */
 
  /*--- GE multi-echo EPI sequences [RCR - Aug 2014] ---*/
- "0008 0018" ,  /* SOP (service object pair) Instance UID - unique to DICOM file*/
- "0018 0081" ,  /* ACQ Echo Time    (new GE field) [RCR - 24 Mar 2022] */
- "0018 0086" ,  /* ACQ Echo Number  (new GE field) [RCR - 24 Mar 2022] */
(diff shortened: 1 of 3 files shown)


jberrebi:
Sorry, I don't know how I managed to load the version with an upper case A in the  dicom tag "0008 002A" in mri_dicom_elist.h. The code with that upper case A might compile but it will fail when running real-time. I really thought I changed to lower case "a" before uploading to github, but apparently I picked up the wrong version or there was something left in some cache.
It became "A" instead of "a" when I was cleaning the code from comments in french and probes with ai.
Dicom tags should be in lower case here. then real-time works.

afni-rickr:
Thanks for that.  I have not yet gotten to review and test this, but it's great to see.

PR #942, opened 2026-08-27
CMakeLists_x_dependent.txt: fix spelling mistake in target

mrneont:
@pmolfese @leej3 @afni-rickr : am I correct that this PR fixes a typo in the CMake targets? or was the initial spelling purposeful to overcome a typo somewhere?

### DIFF
--- a/src/CMakeLists_x_dependent.txt
+++ b/src/CMakeLists_x_dependent.txt
@@ -33,7 +33,7 @@ target_link_libraries(
            X11::X11
            X11::Xext
   PRIVATE
-          $<IF:$<BOOL:${USE_SYSTEM_XMHTML}>,XmTHML::XmHTML,AFNI::XmHTML>
+          $<IF:$<BOOL:${USE_SYSTEM_XMHTML}>,XmHTML::XmHTML,AFNI::XmHTML>
 )
 
 set_source_files_properties(


leej3:
Thanks. Just a silly typo on my part. The practical effect is that, when USE_SYSTEM_XMHTML=ON, CMake fails during generation with target was not found. With the option disabled, the bundled AFNI::XmHTML target is selected, so the typo remains hidden.
CMake’s dependency model is based around find_package(). Find modules locate dependencies and define the imported targets they export for downstream users. CMake provides many modules, and they continue to improve across releases; for less common dependencies, writing a project-specific module is the practical approach, as we have done for XmHTML. That module exports [these targets](https://github.com/afni/afni/blob/master/cmake/FindXmHTML.cmake#L70-L85), including XmHTML::XmHTML.

One way to catch this class of issue without too much compute is to test configuration and generation without compiling.  A gotcha: CMake caches configuration decisions and generated files in the build directory so reusing one directory while changing provider options can therefore produce misleading results. Although newer CMake versions support --fresh, AFNI’s minimum CMake version predates it, so separate build directories for each configuration are the most portable and least surprising approach.

As a starting point something like:
* core-only
* GUI enabled
* SUMA enabled

Each could be tested with the bundled providers and, where applicable, selected system providers. If it proves useful the testing matrix could be expanded.

mrneont:
@leej3 : cool, thanks, I figured it was just a typing blip that we could update. Thanks for verifying that.

I don't have a quick reply about making deeper structural changes mentioned here.  We will hopefully make some updates across Cmake at some point to simplify structure/dependencies.

leej3:
"making deeper structural changes mentioned here" to be clear I think this would just be 100-200 lines of cmake (or a shell script driving cmake)... but yes having a think about the different configurations and what should be supported would help. I tried to make it as configurable as possible but, as this demonstrates, there are broken combinations that can be lurk. @pmolfese mentioned an alternative strategy to adding targets to a build.

PR #944, opened 2026-08-27
3dReHo: fix tie handling in CalcRanksForReHo

### LINKED ISSUE #945: 3dReHo: tie detection truncates float data to int, biasing Kendall’s W on scaled data (fix available)
## Summary

`CalcRanksForReHo()` in `src/ptaylor/rsfc.c` computes ranks for Kendall's W from a correct float sort, but performs **tie detection on values truncated to `int`**:

```c
int *sorted=NULL; // hold sorted time course, assume has been turned into int
...
sorted[m] = THD_get_voxel(T,idx, gsl_permutation_get(P,m));   /* float -> int truncation */
...
if( (sorted[m]==sorted[m-1]) && LENTIE==0 ) { ... }           /* tie test on truncated ints */
```

Any two distinct float values sharing the same integer part are falsely treated as tied: their correct ranks are overwritten with the group-average rank, and the tie-correction term Σ(t³−t) in the W denominator is spuriously inflated.

## Why this matters

The bias depends entirely on the numeric scale of the input:

- **Raw-intensity EPI** (values in the hundreds/thousands): essentially unaffected — presumably why this went unnoticed.
- **Percent-signal-change or unit-variance data** (values of order 1 — e.g. `afni_proc.py` errts output, or z-scored series): heavily affected.

Simulating the exact code path (M=27 neighbors, N=150 time points, ρ=0.3) against a correct tie-corrected Kendall's W gives a relative error in W of about **−1% at data scale ~3, −19.5% at scale ~1, −58.8% at scale ~0.5**.

Because the bias tracks the *local* signal amplitude, ReHo maps computed from scaled data are distorted rather than uniformly rescaled — the error varies voxel-to-voxel and subject-to-subject, and can confound ReHo group differences with amplitude (ALFF-like) differences. The `-chi_sq` output inherits the bias. Published methods sections suggest running `3dReHo` on scaled or nuisance-regressed residual data is a common configuration (it is also the configuration XCP-D uses for volumetric ReHo).

## Verification

The patched tie-adjustment logic was extracted into a standalone harness (GSL index-sort emulated with a stable qsort) and compared against an independent midrank + tie-count reference:

- random float data at scale ~1 and ~1000: ranks and NTIE match the reference exactly;
- integer data with genuine ties, including a tie group at the series maximum: exact match;
- an all-constant series (one whole-series tie): exact match;
- 2,000 fuzz trials of length-200 series drawn from a half-integer grid with heavy random ties: all ranks and tie counts match exactly.

Genuine exact ties (integer-valued input) are still handled identically after the fix.

cindykrafft:
Fixes #945.

CalcRanksForReHo declares its working array as

```c
int *sorted=NULL; // hold sorted time course, assume has been turned into int
```

but fills it from `THD_get_voxel()`, which returns `float`. Every value is truncated toward zero before the tie test, so two distinct values sharing an integer part are treated as tied — their ranks get replaced by an average rank, and the Kendall's W tie-correction term is inflated. The comment says "assume has been turned into int", but I couldn't find a path where that's guaranteed for the input time series.

The severity scales with signal amplitude, so it biases ReHo differently across datasets. Comparing the current rank computation against a midrank reference over 46-point series, 20k trials per condition:

| input scale | series with incorrect ranks | max rank error |
|---|---|---|
| values in [0, 1) — e.g. percent signal change, unit-variance scaling | 100% | 22.5 |
| values in [0, 10) | 100% | 7.0 |
| values in [0, 1000) — typical raw EPI | 64% | 1.5 |

On data of order 1 every value collapses into a single tie group, which is the degenerate case.

Storing the value as `float` makes the comparison lossless. I re-verified the merged result — your in-loop tie handling plus this change — against an independent midrank + tie-count reference: 20,000 fuzz trials across heavy-tie integer data, small floats, and one-decimal data, plus targeted cases for ties at the start, at the end, spanning the whole series, and n=2. Ranks and tie counts match exactly.

**One thing to flag on CI:** `test_3dReHo` and `test_3dReHo_with_box` fail here, but they are also failing on `master` — since c4515647, the merge of the tie fix. The stored reference in `afni_ci_test_data/sample_test_output/ptaylor/` predates that fix, so it encodes the old tie behaviour. Master shows 14.6% of voxels mismatched; this branch shows 14.7%, the small extra being the truncation fix. So the sample output needs regenerating regardless of this PR, and I don't have write access to `afni_ci_test_data` to do it. Happy to generate and send the updated reference if that's useful.

### DIFF
--- a/src/ptaylor/rsfc.c
+++ b/src/ptaylor/rsfc.c
@@ -60,7 +60,7 @@ int CalcRanksForReHo(float *IND, int idx, THD_3dim_dataset *T, int *NTIE,
   int LENTIE = 0;
   float TIERANK;
   int *toP=NULL; // to reset permuts
-  int *sorted=NULL; // hold sorted time course, assume has been turned into int
+  float *sorted=NULL; // hold sorted time course
   int val;
 
   // GSL stuff
@@ -69,7 +69,7 @@ int CalcRanksForReHo(float *IND, int idx, THD_3dim_dataset *T, int *NTIE,
 
 
   toP = (int *)calloc(TDIM,sizeof(int)); 
-  sorted = (int *)calloc(TDIM,sizeof(int)); 
+  sorted = (float *)calloc(TDIM,sizeof(float)); 
 
   if( (toP ==NULL) || (sorted ==NULL) ) { 
     fprintf(stderr, "\n\n MemAlloc failure.\n\n");


mrneont:
Yes, agreed, this had assumed a mapping of time series values to ranks, which had not been added. Thanks!

PR #958, opened 2026-08-31
3dTshift: -no_detrend demeaned the wrong array for the second voxel of each pair

### LINKED ISSUE #957: 3dTshift -no_detrend: second voxel of each pair is shifted without demeaning (copy-paste bug, fix available)
In 3dTshift's paired-voxel loop (`3dTshift.c` ~line 630), the const-detrend branch for the *second* voxel of each pair passes the wrong array:

```c
if( TS_detrend )
   THD_linear_detrend( ntt-ignore , gar+ignore , &g0,&g1 ) ;
else
   THD_const_detrend( ntt-ignore , far+ignore , &g0 ) ;   /* should be gar */
```

With `-no_detrend`, `far` (already demeaned in its own block just above) is demeaned a second time — a near no-op leaving g0 ≈ 0 — while `gar` is never demeaned at all. `gar` is then shifted with its full DC offset in place: the shift kernels' out-of-range handling is zero fill (and Fourier shifting wraps a zero-padded array), both of which assume ~zero-mean input, and afterward only g0 ≈ 0 is re-added. The result is systematic corruption of time points near run edges in every second voxel (odd x-index): values pulled toward the series minimum, or wrap-contaminated with `-Fourier` — even/odd-voxel striping at run edges under `-no_detrend`. Interior time points are unaffected. (The library counterpart `thd_tshift.c` uses `gar` correctly.)

cindykrafft:
Fixes #957 .

One-word fix: pass `gar+ignore` (not `far+ignore`) to `THD_const_detrend` in the second voxel's branch, mirroring the first voxel's block. Previously, with `-no_detrend`, the second voxel of each pair was shifted with its full DC offset against zero-filled (or Fourier-wrapped) edges and only a ≈0 mean was restored afterward, producing even/odd-voxel striping near run edges. `-no_detrend` forces `TS_rlt=2`, so the restore path re-adds exactly `g0` and the linear-trend variables are never read in this mode. Compiles cleanly with `-Wall`.

### DIFF
--- a/src/3dTshift.c
+++ b/src/3dTshift.c
@@ -630,7 +630,10 @@ int main( int argc , char *argv[] )
             if( TS_detrend )
                THD_linear_detrend( ntt-ignore , gar+ignore , &g0,&g1 ) ;
             else
-               THD_const_detrend( ntt-ignore , far+ignore , &g0 ) ;
+               THD_const_detrend( ntt-ignore , gar+ignore , &g0 ) ; /* was
+                    far+ignore: with -no_detrend the mean was removed from
+                    far a second time (a near no-op) while gar was shifted
+                    with its full DC offset against the zero-filled edges */
 
             for( gmin=gmax=gar[ignore],jj=ignore+1 ; jj < ntt ; jj++ ){
                     if( gar[jj] < gmin ) gmin = gar[jj] ;


afni-rickr:
Great, thanks!

PR #962, opened 2026-08-31
3dROIstats: force the Bessel factor into floating point in -sigma/-nzsigma

### LINKED ISSUE #961: 3dROIstats -sigma/-nzsigma: Bessel correction is an integer-division no-op — reported SDs are population SDs (fix available)
In `3dROIstats.c` (~lines 1183 and 1195), the sample-variance factor is computed in `long` arithmetic:

```c
sig = sqrt((voxels[i] / (voxels[i] - 1)) * (sumsq[i] - mean * mean));
```

`voxels[i]/(voxels[i]-1)` equals 1 for every N ≥ 2, so the reported per-ROI SD is the population SD, biased low by √((N−1)/N): −0.5% at N=100, −5.1% at N=10, −29% at N=2. It also makes `3dROIstats -sigma` inconsistent with `3dmaskave -sigma`, which computes the (N−1)-denominator sample SD correctly — the same ROI gives two different "SDs" in the two tools. Same bug in the `-nzsigma` branch.

cindykrafft:
Fixes #961 

`voxels[i] / (voxels[i] - 1)` in `long` arithmetic is 1 for every N ≥ 2, so `-sigma`/`-nzsigma` reported the population SD instead of the intended sample SD (biased low by √((N−1)/N); −5.1% at N=10). Cast the divisor to `double` in both branches, making the output the (N−1)-denominator sample SD and consistent with `3dmaskave -sigma`. Compiles cleanly with `-Wall`.

### DIFF
--- a/src/3dROIstats.c
+++ b/src/3dROIstats.c
@@ -1179,8 +1179,11 @@ int main(int argc, char *argv[])
                      sumsq[i] /= (double) voxels[i];
                      if (voxels[i] == 1)
                         sig = 1e30; /* a really big number */
-                     else
-                        sig = sqrt((voxels[i] / (voxels[i] - 1)) * (sumsq[i] - mean * mean));
+                     else  /* force the Bessel factor N/(N-1) into floating
+                              point: in long arithmetic it was always 1, so
+                              the reported SD was the population SD, biased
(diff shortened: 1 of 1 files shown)


afni-rickr:
Thanks!

PR #976, opened 2026-09-05
Fix incompatible pointer types build error

ghisvail:
AFNI fails to build with `-Wincompatible-pointer-types` due to a unint32_t / uint64_t pointer type mismatch.


mrneont:
Ah, thanks, those are correct updates to the types, thanks.

PR #980, opened 2026-09-16
C23

afni-rickr:
Apply commits from @berndf.  Update typedef for older compilers.


afni-rickr:
Hi @berndf.  I just wanted to add a minor update for the forward struct declarations on older systems.  Does this seem reasonable?  I think this code looks good.  Thanks!

berndf:
> Hi @berndf. I just wanted to add a minor update for the forward struct declarations on older systems. Does this seem reasonable? I think this code looks good. Thanks!

Sure, didn't know that typedef struct is not available. Should work!

Thanks for merging and sorry that there seems to be one showstopper on macos. I think it is due to one of the patches that I restricted to Linux but also needed by macos...

afni-rickr:
Using typedef is okay, but pre-C11 compilers do not like:
```
typedef struct A A;
typedef struct A { defining ... } A;
```
apparently because it looks like multiple typedefs for A.

I don't think your changes are causing the testing failure.  We have some cleaning to do there.
But is there another update you would like to make regarding that Linux restriction?

Thanks!

berndf:
> I don't think your changes are causing the testing failure. We have some cleaning to do there. But is there another update you would like to make regarding that Linux restriction?

I checked again - I interpreted the failure of the macos CI build as caused by my changes and thought that it was due to change "In mpeg_encodedir/headers/libpnmrw.h, do use malloc.h instead of trying to use a K&R style own declaration (presently only if LINUX2 is defined, should probably by used on all modern systems)." being only used on Linux and probably macos needing it too. However, the macos build fails during setup, not compilation. Not having access to macos for testing, I don't have other updates regarding this.
The newer compilers are quite good at warning if provided buffer sizes are potentially too small to hold formatted strings. I'd file a new pull request once I get around to fix more warnings.

Thanks and best regards!

PR #985, opened 2026-09-19
Makefile.INCLUDE: dependence related to prog_opts and apsearch; start…

mrneont:
… to edit thd_getpathprogs.c


mrneont:
This successfully fixes the apsearch behavior that got broken by a recent update to prog_opts.c;  well, more correctly speaking, apparently that update revealed an underlying break elsewhere. This should be more robustly resolved now.

